PR #347, opened 2014-10-30
Completion handler exception handling

jonlipsky:
Added exception handling to all of the completion handlers.  Without the exception handling, if an exception occurs within the handlers, the app will crash on Xamarin.iOS since it's being run on a new thread.


zgramana:
Pretty sure this is fixed in couchbase/couchbase-lite-net#348, which threw an exception in `Stopping` when `Task.WhenAll` had a task get cancelled.

Issue #364, opened 2014-12-10
Manager static ctor crashes on Azure


krcourville:
Behavior was observed in a Web Api app deployed to an Azure web role.  When creating an instance of Manager using the ctor that takes a DirectoryInfo, the following exception is thrown:

`System.ArgumentException: "The path is not of a legal form"  

This is occurring in the state constructor of the Manager class as a result of an Azure bug and/or environment difference. More info here: https://nuget.codeplex.com/workitem/3415

Here's my proposed fix in Manager.cs:

        static Manager()
        {
            illegalCharactersPattern = new Regex(IllegalCharacters);
            mapper = new ObjectWriter();
            DefaultOptions = ManagerOptions.Default;
            //
            // Note: Environment.SpecialFolder.LocalApplicationData returns null on Azure (and possibly other Windows Server environments)
            // and this is only needed by the default constructor or when accessing the SharedInstanced
            // So, let's only set it only when GetFolderPath returns something and allow the directory to be
            // manually specified via the ctor that accepts a DirectoryInfo
            var defaultDirectoryPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
            if (!String.IsNullOrWhiteSpace(defaultDirectoryPath))
            {
                defaultDirectory = new DirectoryInfo(defaultDirectoryPath);
            }
        }

PR #365, opened 2015-01-01
Fix: Manager static ctor crashes on Azure #364

krcourville:
This is my official first pull request on an open source project, so let me know if I'm doing something stupid.

This patch is for #364 which allows cblite to function in a .NET web app hosted in Azure.  By enabling the .NET API to working in Azure, we are able to much more quickly generate initial snapshots of databases for mobile clients (as opposed to waiting for hundreds of thousands of records to stream down).


zgramana:
Awesome! Thanks @krcourville!

PR #393, opened 2015-03-16
Fix null reference exception when setting cookie

elevine:
This fixes a null reference exception I encountered when using CBL from
my own .Net45 unit tests.  I was getting a null reference exception when
setting the replicator cookies.  The CBL unit test for adding/deleting
cookies seems to work fine without this though.

### DIFF
--- a/src/Couchbase.Lite.Shared/Util/CookieStore.cs
+++ b/src/Couchbase.Lite.Shared/Util/CookieStore.cs
@@ -198,7 +198,7 @@ namespace Couchbase.Lite.Util
                 var json = reader.ReadToEnd();
 
                 var cookies = JsonConvert.DeserializeObject<List<Cookie>>(json);
-				cookies = cookies ?? new List<Cookie>();
+                cookies = cookies ?? new List<Cookie>();
 
                 foreach(Cookie cookie in cookies)
                 {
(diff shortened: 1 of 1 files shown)


zgramana:
Thanks, @elevine! Looks great. Just two things:
 * Tabs > Spaces :)
 * Sign the [New Contributor Agreement](http://review.couchbase.org/#/settings/agreements)

elevine:
@zgramana I signed the agreement.  Do you want me to change it to spaces? We use tabs at work, so I've got VS set to that by default.  Tabs vs Spaces is one battle I choose to stay out of :)

zgramana:
@elevine thanks! And yeah, please change them since everything else is spaces, your code is the oddball. ;) We've setup the Xamarin Studio project preferences to use spaces, but we'll try to do the same for VS so people don't need to think about it.

elevine:
@zgramana The spacing should be fixed with the commits I added.  Let me know otherwise.

zgramana:
Spacing looks good.

Your next commit changed the version of NUnit and made some other csproj rewrite changes. Can you please back those out?

elevine:
Sorry about that, it should be all good with commit 50ceb3d.

zgramana:
No problem. I've only done it about a thousand times myself. :)

zgramana:
And thanks again for the PR.

PR #394, opened 2015-03-31
Users can control the max number of HTTP Connections and sizes of bulk downloads

zippo227:
@brettharrison had to take it upon himself to reduce the maximum size that Couchbase Lite tries to download at once. When users are on bad connections, the bulk downloads would fail every time because they were too large. He also noticed several issues with concurrent download threads, so he created a parameter to reduce the number of maximum http connections to allow at once.

I fixed several places where the dll would crash hard because things were not null checked. Other pieces of code that could throw exceptions were for some reason being performed outside of the try catch. We request you pull these changes into the main repository to benefit others and move this dll along.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ TestResults
 *.user
 *.sln.docstates
 local-test.properties
+*.DS_Store
 
 # Build results
 [Dd]ebug/
@@ -37,7 +38,7 @@ x64/
 *.log
 *.vspscc
 *.vssscc
-.builds
+.build
 
 # Package
 *.xam
--- a/samples/CouchbaseSample.iOS/CouchbaseSample.iOS.csproj
+++ b/samples/CouchbaseSample.iOS/CouchbaseSample.iOS.csproj
@@ -99,20 +99,20 @@
     <MtouchArch>ARMv7</MtouchArch>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Newtonsoft.Json">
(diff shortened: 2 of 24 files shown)


zgramana:
Thanks for the pull request! We will review this this week, but before we can accept, we need @brettharrison and Andy Martin to sing the [New Contributor Agreement](http://review.couchbase.org/#/settings/agreements).

borrrden:
This pull request cannot be accepted as it is, since it undoes a lot of .NET 3.5 compatibility stuff.  I'll work on it a bit when I get the chance.

zippo227:
@borrrden the intention was not to break that compatibility. Hopefully those side-affects can be easily remedied. We are targetting .NET 4.5, so I can imagine something haven broken for the earlier .NET versions. @brettharrison and I are in process of getting approval to sign the contribution agreement.

borrrden:
Well, for example there are some places where lines containined #if NET_3_5 have been deleted.  If you could undo that kind of stuff on your end it would help the process move along faster :).  Otherwise I will do it sometime this week.

zippo227:
@borrrden I did a pass through the files and reverted the changes you suggested. I'm sure you will also need to do a pass through, but I took care of the things that stuck out to me as unnecessarily different between origin and our fork. I also fixed a case where your master was not using the StringEx.

zippo227:
@zgramana we've both submitted the Contributor Agreement.

borrrden:
Well I merged things according to the GitHub instructions but I guess because I rebased it didn't detect the change, so I had to manually close the PR.  It is merged though, with some stylistic changes.

zippo227:
@borrrden I'm very glad to have our contributions in the official repository.

PR #399, opened 2015-04-08
Setting Pusher.DocIds will restrict the docs actually pushed accordingly

### LINKED ISSUE #397: Setting Pusher.DocIds does not filter which docs are actually pushed
Given:
```csharp
pusher.DocIds = new[] { "foo" };
```
Expected:
Only the `foo` doc gets pushed.

Observed:
All unpunished docs are pushed.

zgramana:
Fixes #397 thought it still needs unit tests.


borrrden:
I will merge this after the unit tests are finished

PR #403, opened 2015-04-21
Unity 4.6 compatible logging

jonghough:
Current logging functions LogFormat, LogWarningFormat, LogErrorFormat, are only available from Unity 5.0. This gives cross compile errors when trying to build for iOS, if Unity version is 4.x. Since many projects are still using Unity 4.x, it seems better to use the original logging API.

### DIFF
--- a/src/Couchbase.Lite.Unity/UnityLogger.cs
+++ b/src/Couchbase.Lite.Unity/UnityLogger.cs
@@ -50,10 +50,10 @@ namespace Couchbase.Lite.Unity
         {
 
         }
-		// Unity 4.x API compatible Unity Logging.
-		public void LogUnity(string log){UnityEngine.Debug.Log(log);}
-		public void LogWarningUnity(string log){UnityEngine.Debug.LogWarning(log);}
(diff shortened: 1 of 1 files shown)


borrrden:
Oh, I didn't know that!  By the way can you change your tabs to spaces and sign the [New Contributor Agreement](http://review.couchbase.org/#/settings/agreements) (If you haven't done so in the past).

jonghough:
I changed tabs to spaces in my previous edits.
By the way, it seems Google no longer supports OpenID 2.0 (as of April 20), so I cannot use my gmail account to log in, which is also my Github account's email address.

borrrden:
I've passed along that message, so hold that thought for a bit.

borrrden:
It looks like you are going to have to log in using another OpenID Provider (Yahoo is probably the easiest).  You can choose which email address you assign to the agreement after login.  Sorry for the trouble.

jonghough:
No trouble. I logged in using a different email account, and I signed the Agreement, registering my gmail account as the preferred email account. (username JonHough).

PR #413, opened 2015-04-28
Protect the sqlite raw dictionary with a lock to prevent issues during multithread access

brettharrisonzya:
This is a workaround to protect the Dictionary in SQLitePCL.raw until it can be fixed in that library.  See https://github.com/ericsink/SQLitePCL.raw/issues/45 for a full issue description.  If SQLitePCL.raw gets updated to handle SQLITE_OPEN_FULLMUTEX mode this lock can be removed.


borrrden:
I'll be backing out the "added a version file" commit at least since this file is set up to be generated before each build (custom commands were a bit flaky so I changed to the msbuild community extensions).  As for the rest, watch out for tabs vs spaces ;).  Will have another look at this tomorrow.

brettharrisonzya:
Fixed the tabs and spaces.  Deleted the version file.

brettharrisonzya:
Any thoughts of these changes?

borrrden:
I haven't had a whole lot of time to review this yet.  One part that concerns me is that there are places where the sqlite statements were being disposed before (begin transaction, etc) and now they are not being disposed.  Is that on purpose?  Also, Android doesn't work with the newest JSON.NET for whatever reason so I'll keep that one back on 6.0.4.

brettharrisonzya:
Good catch on the Dispose.  I put the "using" blocks back in.
My code puts all "prepare" calls through the BuildCommand which calls raw.sqlite3_prepare_v2 with a lock on it.

The lock is also used on the Dispose of sqlite3 statements.  This protects the Dictionary object internal to SQLitePCL.raw while locking the minimum amount of code.  If SQLitePCL.raw adds support for multithreading then you can later remove the lock call and lock object by just removing a few lines of code.

Issue #414, opened 2015-04-30
NullReferenceException when pulling a document with _deleted set to null


elevine:
Through the Sync Gateway's REST API, our code was setting the _deleted property in the document to null.  This would cause an exception in Couchbase Lite when it tried to sync the document.  It seems to be caused by the way _deleted is being extracted by this constructor in the RevisionInternal class:

```internal RevisionInternal(Body body)
            : this((string)body.GetPropertyForKey("_id"), (string)body.GetPropertyForKey("_rev"), (body.HasValueForKey("_deleted") && (bool)body.GetPropertyForKey("_deleted")))
        {
            this.body = body;
        }
```

Here is the stack trace:
```
Replication: SendAsyncMultipartDownloaderRequest has an error occurred.
04-30 15:15:35.645 I/mono-stdout( 1829): Replication: SendAsyncMultipartDownloaderRequest has an error occurred.

    System.NullReferenceException: Object reference not set to an instance of an object
      at Couchbase.Lite.Internal.RevisionInternal..ctor (Couchbase.Lite.Internal.Body body, Couchbase.Lite.Database database) [0x00000] in <filename unknown>:0 
      at Couchbase.Lite.Internal.RevisionInternal..ctor (IDictionary`2 properties, Couchbase.Lite.Database database) [0x00000] in <filename unknown>:0 
      at Couchbase.Lite.Replicator.PulledRevision..ctor (IDictionary`2 properties, Couchbase.Lite.Database database) [0x00000] in <filename unknown>:0 
      at Couchbase.Lite.Replicator.Puller+<>c__DisplayClass15.<PullRemoteRevision>b__14 (System.Object result, System.Exception e) [0x00000] in <filename unknown>:0 
      at Couchbase.Lite.Replication+<>c__DisplayClass1f.<SendAsyncMultipartDownloaderRequest>b__1c (System.Threading.Tasks.Task`1 responseMessage) [0x00000] in <filename unknown>:0 
```

PR #415, opened 2015-05-01
Added bool type checking to the RevisionInternal constructor.  

elevine:
This handles the case where _deleted is set to either null or a non-boolean type in the JSON.  Addresses issue #414.


borrrden:
Well there is no harm in this, but just as a side note you shouldn't be setting the _deleted property manually.  In fact I think I will modify this later to use the GetCast function that I introduced.

elevine:
@borrrden I ran into this because a document on the server had _deleted set to null, which caused the replicator to stop working,  I'm sure the code can be refactored to something better then I submitted, but I thought it was important to make it a little more robust.

PR #417, opened 2015-05-11
Release/1.1

ricardonns:


borrrden:
release/1.1 will not be merged into master as-is.  I'm cherry picking commits into master as I go along so I don't insert improper ones (like changing the builds to "official", etc)

PR #420, opened 2015-05-14
Implement Full Text Search

ylatuya:
Initial implementation of Full Text Search based on the [iOS implementation](https://github.com/couchbase/couchbase-lite-ios/wiki/Full-Text-Search)

The query property fullTextSnippets is not implemented yet.

The results does not include matchCount, textRangeOfMatch, termIndexOfMatch and snippetWithWordStart:wordEnd: yet.

### DIFF
--- a/src/Couchbase.Lite.Shared/Database.cs
+++ b/src/Couchbase.Lite.Shared/Database.cs
@@ -793,7 +793,7 @@ CREATE TABLE maps (
   view_id INTEGER NOT NULL REFERENCES views(view_id) ON DELETE CASCADE, 
   sequence INTEGER NOT NULL REFERENCES revs(sequence) ON DELETE CASCADE, 
   key TEXT NOT NULL COLLATE JSON, 
-  value TEXT); 
+  value TEXT);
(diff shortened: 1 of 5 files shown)


zgramana:
This will cause crashes in older copies of SQLite that don't have FTS4 support. That includes many common OS versions on many popular platforms. In fact, iOS only just got in 8.2 IIRC. We have to ship custom SQLite binaries for each platform in order to merge this. Also, keep this in mind for your own application.

ylatuya:
Correct, we do ship our custom sqlite library with fts4 support but I wasn't considering others.

Plus the migration path is not correct now that I double check, because it would need to be done in a new database version instead of the old v7

borrrden:
In fact I think we are going to start bundling the sqlite library soon though (for SQLCipher support etc) so this is something I will look into at that time

borrrden:
@ylatuya This code is too old to merge now, but I am working on implementing full text search for 1.5 (https://github.com/couchbase/couchbase-lite-net/issues/647).  I'm going to close this PR though

PR #422, opened 2015-05-20
MD5 hash wasn't being checked.

jonlipsky:
It should have been checking for the MD5 hash as a fallback as opposed to checking for the SHA1 hash twice.


borrrden:
Oh dear!  I'll get this in quickly.

PR #471, opened 2015-07-30
Feature/defered attachments

brettharrisonzya:
New feature to defer downloading attachments during pull replication.  

Attachments can be downloaded later on demand by using GetDeferedAttachment instead of GetAttachment.  GetDeferedAttachment will attempt to find the attachment in the local database and if not found download if from the sync gateway and then store it locally.

To activate this feature set:
ManagerOptions.Default.DownloadAttachmentsOnSync = false;

Then use GetDeferedAttachment instead of GetAttachment and be sure to handle null attachment bodies in the event the attachment cannot be downloaded if the network is unavailable.

Potential uses include having different image (etc...) assets for different phone models and then the app will only download the assets that if needs.

This is related to https://github.com/couchbase/couchbase-lite-ios/issues/79


brettharrisonzya:
Added code to merge http attachment requests when attachment is already queued to be downloaded.

PR #494, opened 2015-08-20
Fix for issue 492 iOS JSON deserialization can't go directly to an IL…

z000z:
Fix for issue 492 iOS JSON deserialization can't go directly to an IList<T> have to construct the list from the inner types manually.


borrrden:
I prefer to put this fix into the unity-ga branch rather than the issue/481 branch since it is not specific to that issue.  I have done that now, so I will close this PR.

PR #503, opened 2015-09-09
Url issue

jonlipsky:
Fixed an issue doing reachability test with basic authentication.
Fixed a possible (based on timing) null pointer exception starting the replication process.

### DIFF
--- a/src/Couchbase.Lite.Shared/Manager/NetworkReachabilityManager.cs
+++ b/src/Couchbase.Lite.Shared/Manager/NetworkReachabilityManager.cs
@@ -48,15 +48,13 @@ namespace Couchbase.Lite
 
             var uri = new Uri (remoteUri);
             var credentials = uri.UserInfo;
-            if (credentials != null)
-            {
+            if (credentials != null) {
                 remoteUri = string.Format ("{0}://{1}{2}", uri.Scheme, uri.Authority, uri.PathAndQuery);
                 request = WebRequest.CreateHttp (remoteUri);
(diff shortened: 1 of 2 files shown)


borrrden:
Hey nice to hear from you again!  Would you mind telling me what the issue was with the basic auth?  I ran into this problem before and corrected it by adding lines 72 - 75.  If the authorization is incorrect usually it is returned as a WebException, but I don't really care that it did that because it still means the URL is reachable.  So I checked for a WebException and still return true if it is a protocol error.  Has this stopped working for you?

jonlipsky:
Hi Jim, yes, it stopped working for me with the latest versions of
Xamarin.iOS.  I had to add this code to detect and use basic authentication
to get it to work, otherwise the request would simply timeout.

borrrden:
That's weird, is that normal CouchDB behavior?  The normal behavior for an HTTP server is to return a 401 response.  I'm worried about this code because basic auth is not the only form of authentication that could pass through here (digest auth is on the way as well).

jonlipsky:
CouchDB returns a 401 like you'd expect.  Everything was working fine until
recently, and I think it has to do with Xamarin.  I understand if you don't
want to  accept the pull request, as I'm used to having  to use a slightly
modified version of couchbase lite anyway.  I haven't yet been able to use
a release version in my app without running into some type of issue.

borrrden:
Well, after some research I think that UserInfo only applies to basic auth.  Digest auth doesn't specify the credentials in the URL since it is challenge response so I don't think there is a problem using this as-is.

PR #515, opened 2015-10-08
Fix querying for keys which are arrays

chrisdurtschi:
When querying a map-only view for a specific set of keys, where the keys of the view are arrays, no results are being returned.

The problem occurs when sorting the results in the same order as the keys given to the query. The results are grouped together by key by adding them to a `Dictionary<object, List<QueryRow>>`. However, keys which are arrays are returned as `JArray`, which needs to use `JTokenEqualityComparer` to be used as a `Dictionary` key.

This solution simply converts the keys to JSON strings before adding to and retrieving from the `Dictionary`.

The test was taken from the Android project:
https://github.com/couchbase/couchbase-lite-android/tree/36b977c7389aaebf466229b198784363d7b7cef9/src/androidTest/java/com/couchbase/lite/ViewsTest.java#L583

### DIFF
--- a/src/Couchbase.Lite.Shared/Store/SqliteViewStore.cs
+++ b/src/Couchbase.Lite.Shared/Store/SqliteViewStore.cs
@@ -958,11 +958,12 @@ namespace Couchbase.Lite.Store
             // If given keys, sort the output into that order, and add entries for missing keys:
             if (options.Keys != null) {
                 // Group rows by key:
-                var rowsByKey = new Dictionary<object, List<QueryRow>>();
+                var rowsByKey = new Dictionary<string, List<QueryRow>>();
                 foreach (var row in rows) {
-                    var dictRows = rowsByKey.Get(row.Key);
+                    var key = ToJSONString(row.Key);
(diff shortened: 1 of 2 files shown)


borrrden:
Thanks for this insight!  This is the symptom of a larger problem that I am dealing with.  I'm trying to abstract away JArray and all Newtonsoft classes from the API but there are still some leaks, as you've found.  This solution might work in the interim but I'd also like to try to plug the leak first to see if that solves the problem on its own.  Either way thank you for bringing this to light!

chrisdurtschi:
Sounds good, thanks for your hard work!

borrrden:
There has been some debate internally on whether or not it would be wise or performant to do what I suggested above.  I have been able to do it, but I haven't measured the performance hit yet.  I'm almost done with the current big thing I've been working on and after that I will have time to go back to all the other waiting issues.

borrrden:
I'm going to merge this so that the problem doesn't stay active while we are all racking our brains over here about how to handle the abstraction leak without introducing too much overhead.

PR #530, opened 2015-11-20
Fix deadlock when multiple UpdateIndex operations are queued

jonfunkhouser:
My project makes heavy use of asynchronous queries, and I was occasionally seeing a deadlock of the Manager's workExecutor. 

I did some digging and found that the UpdateJob task continuation was not always being executed, even though the task ran to completion. If another UpdateIndex operation was queued and the previous continuation did not execute, the View would end up waiting forever for the new operation to complete. 

Scheduling the continuation on TaskScheduler.Default ensures the continuation is always executed.

### DIFF
--- a/src/Couchbase.Lite.Shared/View/UpdateJob.cs
+++ b/src/Couchbase.Lite.Shared/View/UpdateJob.cs
@@ -73,7 +73,7 @@ namespace Couchbase.Lite.Internal
                     if(Finished != null) {
                         Finished(this, null);
                     }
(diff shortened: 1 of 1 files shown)


borrrden:
Thanks for this.  It makes sense and I feel silly asking for a signature for this one little thing but, have you signed the New Contributor Agreement at some point in the past?

jonfunkhouser:
Yes, I signed it before submitting this pull request.

PR #555, opened 2016-01-23
AndroidSimpleSync handle deletes

sethrosetter:


sethrosetter:
@borrrden This handles the issue from the sample app side, but you said it may have to do with the internal implementation (get doc id and rev id from enumerator rather than docRevision).

Issue #749, opened 2016-09-23
Attachments failed to replicate when deleted and recreated


borrrden:
https://github.com/couchbase/couchbase-lite-ios/issues/1440

sethrosetter:
Please let me know when this is resolved and I will re-enable the test - https://github.com/couchbaselabs/mobile-testkit/issues/757

PR #780, opened 2016-11-28
Fix for #749

borrrden:
Need to return a proper stub (it is missing a few fields which causes replication to fail)

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/780)

### DIFF
--- a/src/Couchbase.Lite.Shared/Database.cs
+++ b/src/Couchbase.Lite.Shared/Database.cs
@@ -1602,7 +1602,12 @@ namespace Couchbase.Lite
                 var revPos = attachment.GetCast<long>("revpos");
                 if(revPos < minRevPos && revPos != 0) {
                     //Stub:
-                    return new Dictionary<string, object> { { "stub", true }, { "revpos", revPos } };
+                    return new NonNullDictionary<string, object> {
+                        ["stub"] = true,
+                        ["revpos"] = revPos,
+                        ["digest"] = attachment.GetCast<string>("digest"),
+                        ["length"] = attachment.GetCast<long>("length")
+                    };
                 }
 
                 var expanded = new Dictionary<string, object>(attachment);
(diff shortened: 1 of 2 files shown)


src/Couchbase.Lite.Shared/Database.cs
```diff
@@ -1602,7 +1602,12 @@ internal IList<RevisionID> GetRevisionHistory(RevisionInternal rev, IList<Revisi
                 var revPos = attachment.GetCast<long>("revpos");
                 if(revPos < minRevPos && revPos != 0) {
                     //Stub:
-                    return new Dictionary<string, object> { { "stub", true }, { "revpos", revPos } };
+                    return new NonNullDictionary<string, object> {
+                        ["stub"] = true,
+                        ["revpos"] = revPos,
+                        ["digest"] = attachment.GetCast<string>("digest"),
+                        ["length"] = attachment.GetCast<long>("length")
```
snej: Do "digest" and "length" need to be cast? Why not just pass them through?

src/Couchbase.Lite.Shared/Database.cs
```diff
@@ -1769,7 +1774,12 @@ internal bool ProcessAttachmentsForRevision(RevisionInternal rev, IList<Revision
         internal IDictionary<string, object> GetAttachmentsFromDoc(string docId, RevisionID revId)
         {
             var rev = new RevisionInternal(docId, revId, false);
-            LoadRevisionBody(rev);
+            try {
+                LoadRevisionBody(rev);
+            } catch(Exception) {
+                return null;
```
snej: Ignoring exceptions can be a code smell … what's the reason for doing so here?

borrrden:
This pull request is invalid upon further research, closing.

PR #790, opened 2016-12-20
Fixes #749

### LINKED ISSUE #749: Attachments failed to replicate when deleted and recreated

borrrden:
Delay encoding the JSON until after we know whether or not this insertion is on top of a deletion.  On ForestDB this is awkward and requires an extra read, unless there is a better way that I missed.

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/790)

### DIFF
--- a/src/StorageEngines/ForestDB/storage.forestdb.shared/src/ForestDBCouchStore.cs
+++ b/src/StorageEngines/ForestDB/storage.forestdb.shared/src/ForestDBCouchStore.cs
@@ -1314,13 +1314,6 @@ namespace Couchbase.Lite.Storage.ForestDB
                     "Attempting to write to a readonly database (PutRevision)");
             }
 
-            var json = default(string);
-            if(properties != null) {
-                json = Manager.GetObjectMapper().WriteValueAsString(Database.StripDocumentJSON(properties), true);
-            } else {
-                json = "{}";
-            }
-
             if(inDocId == null) {
                 inDocId = Misc.CreateGUID();
             }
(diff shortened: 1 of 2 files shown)


borrrden:
This change will need to be applied to iOS and Java if approved @hideki @pasin

src/StorageEngines/ForestDB/storage.forestdb.shared/src/ForestDBCouchStore.cs
```diff
@@ -1260,8 +1253,35 @@ public void SetInfo(string key, string info)
...
+
+                    // https://github.com/couchbase/couchbase-lite-net/issues/749
+                    // Need to ensure revpos is correct for a revision inserted on top
+                    // of a deletion
+                    var existing = (C4Document*)ForestDBBridge.Check(err => Native.c4doc_getForPut(Forest, docId, prevRevId?.ToString(), deleting,
+                        allowConflict, err));
+
+                    if(existing->IsDeleted) {
+                        var attachments = properties.CblAttachments();
+                        if(attachments != null) {
```
snej: Since this test is cheap and is likely to fail (most docs don't have attachments), it would be best to do it before loading the old revision (i.e. before line 1263.) That will minimize the performance hit.

borrrden:
Added the optimization, is everything else good to go?

borrrden:
@snej Please have a look at this.  I'm planning to merge this on Monday.

borrrden:
Judging by the review of 1440 downstream I will consider this reviewed.

PR #794, opened 2017-01-05
Fixes #792

### LINKED ISSUE #792: P2P 'server' crashes with large changes set on pull
The reported case was with 1.7 million revisions.

borrrden:
Fixes #792


borrrden:
@hideki @pasin For porting reference (mostly in the first two commits)

PR #799, opened 2017-01-10
Added support to use default system proxy

unkani:
If a user has a proxy, TcpClient() by default doesn't take that into account.

Same with the WebSocket class in websocket-sharp, which I believe actually uses TcpClient(), but they have a way to set a proxy.

I'm on .NET version 4.5, but according to documentation, the properties used, such as WebRequest.DefaultWebProxy, Uri.DnsSafeHost has been available since at least .NET 2.0

https://msdn.microsoft.com/en-us/library/system.net.webrequest.defaultwebproxy(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.uri.dnssafehost(v=vs.90).aspx

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/799)

### DIFF
--- a/src/Couchbase.Lite.Shared/Manager/NetworkReachabilityManager.cs
+++ b/src/Couchbase.Lite.Shared/Manager/NetworkReachabilityManager.cs
@@ -51,8 +51,10 @@ namespace Couchbase.Lite
         {
             CouchbaseLiteHttpClientFactory.SetupSslCallback();
             var uri = new Uri(remoteUri);
-            try {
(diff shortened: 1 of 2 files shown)


borrrden:
I can't accept this change quite as it is because of the Ssl stuff.  Tls12 is not a valid enum in .NET 3.5, and so this will need to be configured from the outside I suppose I could base it on the same thing that the non web socket change tracker uses

unkani:
Oh right, I forgot about that!

PR #866, opened 2017-05-28
Added object mapping in Couchbase.Lite

mohsens22:
Dear,
I added a public method in Document class which maps a class to an object.
It works like this : 
 
```
var _doc = new Document();
_doc.MapObject<MyClass>(TheObject);
DB.Save(_doc);
```

Also added a .ToDictionary() extension method in util.
It works like this:

`Object.ToDictionary()`

Due to build problems, It's not fully tested. But it worked in my own UWP project with system.Reflection APIs of UWP.

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/866)

### DIFF
--- a/src/Couchbase.Lite/API/Document/Document.cs
+++ b/src/Couchbase.Lite/API/Document/Document.cs
@@ -27,6 +27,7 @@ using Couchbase.Lite.Logging;
 using LiteCore;
 using LiteCore.Interop;
 using LiteCore.Util;
+using Couchbase.Lite.Util;
 
 namespace Couchbase.Lite
 {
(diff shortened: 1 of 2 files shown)


borrrden:
There are plans for this exact kind of thing but I'm not sure yet of the shape it will take.  It might take this route (with an equivalent for the other way) or it might even pull a class directly out of the database.  I'll ask again when we are going to work on this API, but I can't put in temporary APIs because all mobile teams need to have a consensus first.  So this might sit here for a while, unfortunately.

mohsens22:
OK, Then inform me how do you want to implement this feature, maybe I could do some codings :)

PR #882, opened 2017-06-26
update reference to use the latest CI build

jamiltz:
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/882)


jamiltz:
@borrrden I think we should update the reference to use the CI build, once we know which one it is. That way we can 1. verify the code snippets 2. update the docs 3. release (docs and nuget).

jamiltz:
hmm but then external contributors won't be able to run it without changing the nuget feed and adding the latest db. I take it back for the moment, we should just change it ourselves by hand for the moment and see how that goes.

PR #883, opened 2017-06-26
live query and replication change listener examples.

jamiltz:
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/883)

### DIFF
--- a/docs/examples/api-walkthrough/api-walkthrough/Program.cs
+++ b/docs/examples/api-walkthrough/api-walkthrough/Program.cs
@@ -84,6 +84,17 @@ namespace api_walkthrough
                 Console.WriteLine($"doc ID :: ${row.DocumentID}");
             }
 
+            // live query
+            var liveQuery = query.ToLive();
+            liveQuery.Changed += (sender, e) => {
+                Console.WriteLine($"Number of rows :: {e.Rows.Count}");
+            };
+            liveQuery.Run();
+            var newDoc = new Document();
+            newDoc.Set("type", "user");
+            newDoc.Set("admin", false);
+            database.Save(newDoc);
+
             // fts example
             // insert documents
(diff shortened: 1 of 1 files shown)


jamiltz:
@borrrden I added some details in the second commit. But the replication isn't working (document count on SG is still 0). And the replication status is `Busy`. Can you give it a go?

borrrden:
Replication is working fine on my end (59 rows found in SG and replication completes)

borrrden:
Actually on Mac it does what you say...

borrrden:
Oh the logs say why...you can get around this by using an IP address instead.  This will need to be fixed for db012

borrrden:
Actually this needs to be fixed in the .NET Core runtime itself (and actually it already has).  I tried with .NET Core 2.0 Preview 1 and replication works on Mac.

docs/examples/api-walkthrough/api-walkthrough/Program.cs
```diff
@@ -85,11 +86,11 @@ static void Main(string[] args)
             }
 
             // live query
-            var liveQuery = query.ToLive();
+            var liveQuery = query.ToLiveQuery();
```
jamiltz: For DB11, should this be `.ToLive();`?

docs/examples/api-walkthrough/api-walkthrough/Program.cs
```diff
@@ -85,11 +86,11 @@ static void Main(string[] args)
...
             liveQuery.Changed += (sender, e) => {
                 Console.WriteLine($"Number of rows :: {e.Rows.Count}");
             };
-            liveQuery.Run();
+            liveQuery.Start();
```
jamiltz: For DB11, should this be `.Run();`?

borrrden:
No, db011 is b0154 not b0157

jamiltz:
👍

PR #896, opened 2017-07-27
conflict resolver example.

jamiltz:
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/896)

### DIFF
--- a/docs/examples/api-walkthrough/api-walkthrough/ExampleConflictResolver.cs
+++ b/docs/examples/api-walkthrough/api-walkthrough/ExampleConflictResolver.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using Couchbase.Lite;
+
+
+public class ExampleConflictResolver : IConflictResolver
+{
+    public ExampleConflictResolver()
+    {
+            
+    }
+
+    public ReadOnlyDocument Resolve(Conflict conflict)
+    {
+        var baseProperties = conflict.Base;
+        var mine = conflict.Mine;
+        var theirs = conflict.Theirs;
+
+        return theirs;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


docs/examples/api-walkthrough/api-walkthrough/ExampleConflictResolver.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
borrrden: This file needs a copyright header

docs/examples/api-walkthrough/api-walkthrough/Program.cs
```diff
@@ -120,6 +122,23 @@ static void Main(string[] args)
...
+			 */
+            var theirs = new Document("buzz");
+            theirs.Set("status", "theirs");
+            var mine = new Document("buzz");
+            mine.Set("status", "mine");
+            database.Save(theirs);
+            database.Save(mine);
+
+            var conflictResolverResult = database.GetDocument("buzz");
+            Console.WriteLine($"conflictResolverResult doc.status ::: {conflictResolverResult.GetString("status")}");
```
borrrden: Consider pointing out that the document will have two revisions now

jamiltz:
@borrrden added the copyright header and more description in the conflict resolver example.

mr-jingles:
I have tried to reproduce this - the second Database.Save method will generate a second local revision. However since the two revisions are not conflicting with already existing aerver revisions this will not create a conflict. Both revisions will get pushed to the server, with the second revision as winning revisision. The _changes endpoint (with active_only=true&style=all_docs&include_docs=true') does not show two revisions for this document and the Conflict Resolver also does not get called. This is good as the document is not in a conflict state.

PR #911, opened 2017-10-02
Fixes #910

### LINKED ISSUE #910: Channel removal potentially disrupts attachments
In the following scenario, attachments will not be able to be found:

1. Client goes offline
2. Attachments are added to the document
3. Channel removal happens and the client loses access to the document
4. Client comes online and receives the removal (since there are no attachments on a removal the client assumes there are none at all)
5. Client regains access and discovers that there are attachments, and searches for them.  However it won't find them because it will only check back to the last known revision (the removed one)

borrrden:
Fixes #910 (Channel removal potentially disrupts attachments)

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/911)

### DIFF
--- a/src/Couchbase.Lite.Shared/Database/ICouchStore.cs
+++ b/src/Couchbase.Lite.Shared/Database/ICouchStore.cs
@@ -194,9 +194,11 @@ namespace Couchbase.Lite.Store
         /// <param name="limit">The maximum number of results to return, or if 0, unlimited.</param>
         /// <param name="haveBodies">  On return, if not NULL, then* outHaveBodies will be YES if all the
         /// revisions returned have their JSON bodies available, otherwise NO.</param>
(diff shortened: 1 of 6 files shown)


borrrden:
@snej Since this is a cross platform fix, could you have a look at this implementation of the fix for the attachments issue we discussed?  It's a fairly small PR, so hopefully it is clear what the intent is even from C#.

snej:
It looks good. Just one request: could you make the commit title more descriptive before you merge it? `Fixes #910` means nothing when you're skimming through a list of commits. `Channel removal potentially disrupts attachments` would be better.

borrrden:
Oh oops I meant to be good about that this time but o guess I missed a spot!  I’ll change that commit message.

PR #938, opened 2017-12-21
Fix bug : When a document was created on couchbaseLite than sync that…

yannyann:
… a property was changed twice or more, the replication was failing.

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/938)


yannyann:
I have refactor the code in another branch I will send it to

yannyann:
I think another bug exists I will try to solve it before push the branch

Issue #937, opened 2017-12-21
Pull Sync throws Exception because of updating document with attachment on couchdb


yannyann:
Hello,
I found in the replication.

**Scenario**

- Create a document on couchbaseLite
- Sync on couchdb
- Update this document more that once (ex : twice)
- Sync back
- The replication throws an exception

This bug appears in the method Database.LoadRevisionBody

borrrden:
This affects more than CouchDB it turns out.  There is now another reported instance of this and so it will be scheduled in 1.4.2 instead.

PR #939, opened 2017-12-21
Fix bug #937: When a document was created on couchbaseLite than sync …

yannyann:
…that a property was changed twice or more, the replication was failing.

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/939)


yannyann:
Hello,

This is the version 2 of my solution for the issue #937 (see pull request #938)

Issue #1020, opened 2018-04-26
Closing/Disposing database throws Exception because of unfinished replicators


thomai-d:
See the following code:
![db](https://user-images.githubusercontent.com/24974298/39322063-f305d886-4988-11e8-97c9-7b35c65e072a.png)

The `Close()` call throws an exception if a replicator has been started:

```
Couchbase.Lite.CouchbaseLiteException
  HResult=0x80131500
  Message=CouchbaseLiteException (LiteCoreDomain / 16): Cannot close the database. Please stop all of the replicators before closing the database..
  Source=Couchbase.Lite
  StackTrace:
   at Couchbase.Lite.Database.ThrowIfActiveItems()
   at Couchbase.Lite.Database.<Dispose>b__93_0()
   at Couchbase.Lite.Support.ThreadSafety.DoLocked(Action a)
   at Couchbase.Lite.Database.Dispose()
   at Couchbase.Lite.Database.Close()
   at TestApp.Service.CouchbaseDatabase..ctor() in ...\CouchbaseDatabase.cs:line 75
```

The replicator server is down and the following snippet is the only couchbase code that is executed in my testing application. The exception is raised on Android as well as on UWP.

### Library Version
2.0.0.1

### .NET Runtime
Xamarin Android v27.0.2
Microsoft.NETCore.UniversalWindowsPlatform v6.0.8

### Operating System / Device Details
Android OS 6.01
Windows 10 Version 1709

### Log Output
[log-1524763742.zip](https://github.com/couchbase/couchbase-lite-net/files/1952308/log-1524763742.zip)

### Expected behavior
`this.db.Close()` as well as `this.db.Dispose()` can be called without causing an exception.

### Actual behavior
`this.db.Close()` as well as `this.db.Dispose()` throws an exception when a replicator is configured.

### Steps To Reproduce
Execute given code:
    this.db = new Database(DatabaseName);
    var targetEndpoint = new URLEndpoint(new Uri("ws://192.168.0.11:4984/app"));
    var replConfig = new ReplicatorConfiguration(this.db, targetEndpoint);
    replConfig.Authenticator = new BasicAuthenticator("GUEST", "");
    replConfig.Continuous = true;
    this.replicator = new Replicator(replConfig);
    this.replicator.Start();
    this.replicator.Stop();
    this.replicator.Dispose();
    this.db.Close();

borrrden:
[...]

PR #1030, opened 2018-06-02
#1020 Test to close server when replicator point to none existed uri

Sandychuang8:
Add a test case to test when server is closed when replicator point to none existed uri

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/1030)

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -70,11 +70,12 @@ namespace Test
             ReopenDB();
             Database.Delete("otherdb", Directory);
             _otherDB = OpenDB("otherdb");
-            Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
+            //uncomment the code below when you need to see more detail log
+            //Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
         }
 
         [Fact]
-        public async Task TestCloseServerWhenReplicatorExistsAsync()
+        public async Task TestToCheckReplicatorStopsWhenAttemptToConnectToInvalidEndpoint()
         {
             var targetEndpoint = new URLEndpoint(new Uri("ws://192.168.0.11:4984/app"));
             var config = new ReplicatorConfiguration(Db, targetEndpoint);
@@ -82,19 +83,13 @@ namespace Test
             {
                 repl.Start();
                 var count = 0;
-
-                this.Invoking(x => ReopenDB())
-                    .ShouldThrow<CouchbaseLiteException>(
-                        "because the database cannot be closed while replication is running");
-            
                 while (count++ > 20 && repl.Status.Activity != ReplicatorActivityLevel.Stopped)
                 {
                     WriteLine($"Replication status still {repl.Status.Activity}, waiting for stopped...");
                     await Task.Delay(500);
                 }
 
-                if(repl.Status.Activity == ReplicatorActivityLevel.Stopped)
-                    WriteLine($"Replication status still {repl.Status.Activity}, stopped");
+                count.Should().BeLessThan(20);
             }
         }
 
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -83,7 +83,7 @@ namespace Test
             {
                 repl.Start();
                 var count = 0;
-                while (count++ > 20 && repl.Status.Activity != ReplicatorActivityLevel.Stopped)
+                while (count++ <= 20 && repl.Status.Activity != ReplicatorActivityLevel.Stopped)
                 {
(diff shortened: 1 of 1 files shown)


src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,9 +70,35 @@ public ReplicatorTest()
             ReopenDB();
             Database.Delete("otherdb", Directory);
             _otherDB = OpenDB("otherdb");
+            Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
         }
 
-        #if COUCHBASE_ENTERPRISE
+        [Fact]
+        public async Task TestCloseServerWhenReplicatorExistsAsync()
```
borrrden: The point of the test is to check that the replicator stops when it cannot connect to a non-existent endpoint so I'd change the name to reflect that

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,9 +70,35 @@ public ReplicatorTest()
             ReopenDB();
             Database.Delete("otherdb", Directory);
             _otherDB = OpenDB("otherdb");
+            Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
```
borrrden: This was meant to be temporary.  If left in, all the logs for the rest of the test suite will be at this level

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,9 +70,35 @@ public ReplicatorTest()
...
+            var targetEndpoint = new URLEndpoint(new Uri("ws://192.168.0.11:4984/app"));
+            var config = new ReplicatorConfiguration(Db, targetEndpoint);
+            using (var repl = new Replicator(config))
+            {
+                repl.Start();
+                var count = 0;
+
+                this.Invoking(x => ReopenDB())
+                    .ShouldThrow<CouchbaseLiteException>(
+                        "because the database cannot be closed while replication is running");
```
borrrden: This assertion is not essential to the test, but there is no harm in leaving it in

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,9 +70,35 @@ public ReplicatorTest()
...
+            
+                while (count++ > 20 && repl.Status.Activity != ReplicatorActivityLevel.Stopped)
+                {
+                    WriteLine($"Replication status still {repl.Status.Activity}, waiting for stopped...");
+                    await Task.Delay(500);
+                }
+
+                if(repl.Status.Activity == ReplicatorActivityLevel.Stopped)
+                    WriteLine($"Replication status still {repl.Status.Activity}, stopped");
```
borrrden: This should be an assertion, not just a WriteLine.  The test should fail if the replication does not close.  Note that in the situation we are going to cover, retries will be attempted so it will take quite a while for the replication to stop.  I'll catch up with you about that offline.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,31 +70,26 @@ public ReplicatorTest()
...
             Database.Delete("otherdb", Directory);
             _otherDB = OpenDB("otherdb");
-            Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
+            //uncomment the code below when you need to see more detail log
+            //Database.SetLogLevel(LogDomain.Replicator, LogLevel.Verbose);
         }
 
         [Fact]
-        public async Task TestCloseServerWhenReplicatorExistsAsync()
+        public async Task TestToCheckReplicatorStopsWhenAttemptToConnectToInvalidEndpoint()
```
borrrden: This name is a bit long and I don't think it will show up in the test explorer without making it really wide.  How about `TestReplicatorStopsWhenEndpointInvalid`

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -70,31 +70,26 @@ public ReplicatorTest()
...
-            
                 while (count++ > 20 && repl.Status.Activity != ReplicatorActivityLevel.Stopped)
                 {
                     WriteLine($"Replication status still {repl.Status.Activity}, waiting for stopped...");
                     await Task.Delay(500);
                 }
 
-                if(repl.Status.Activity == ReplicatorActivityLevel.Stopped)
-                    WriteLine($"Replication status still {repl.Status.Activity}, stopped");
+                count.Should().BeLessThan(20);
```
borrrden: Right now the answer is clear as to why the count should be less than 20, but in the future we will probably forget, so it is best to add the optional reason into this assert:

`count.Should().BeLessThan(20, "because otherwise the replicator never stopped");`

This creates a single left to right readable line that explains what the assert is for.

PR #1033, opened 2018-06-05
Register for cancellation callback to detect timeout on replicator connect (#1020)

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -198,20 +198,33 @@ public void Start()
...
                         if (!NetworkTaskSuccessful(t)) {
                             return;
                         }
 
                         _queue.DispatchAsync(StartInternal);
-                    }, cts.Token);
-                } catch (Exception e) {
+
+                        if (tok.IsCancellationRequested)
+                            return;
```
borrrden: The above two lines are unneeded.  Actually it is not possible to make it this far with a cancelled token since it will return false from `NetworkTaskSuccessful`

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -198,20 +198,33 @@ public void Start()
 
                 // STEP 2: Open the socket connection to the remote host
                 var cts = new CancellationTokenSource();
-                cts.CancelAfter(ConnectTimeout);
-                try {
+                var tok = cts.Token;
+                
+                tok.Register(() =>
+                {
+                    _client.Close();
```
borrrden: `tok.Register` returns a value that needs to be disposed here, and the `CancellationTokenSource` itself should also be disposed in here.

borrrden:
I checked out this branch and ran the test with it, but I don't notice any effect.  Let's sync up about this tomorrow.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -197,34 +197,43 @@ public void Start()
...
 
                 // STEP 2: Open the socket connection to the remote host
-                var cts = new CancellationTokenSource();
-                var tok = cts.Token;
-                
-                tok.Register(() =>
-                {
-                    _client.Close();
-                });
+                var cts = new CancellationTokenSource(ConnectTimeout);
```
borrrden: This constructor and the below `CancelAfter` are redundant, they both do the same thing.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -197,34 +197,43 @@ public void Start()
...
+                cts.CancelAfter(ConnectTimeout);
 
+                using (cts.Token.Register(() => { _client.Close(); }))
+                {
+                    _client.Dispose();
```
borrrden: This is not good because it will dispose the client before it is used below (Maybe you meant to put it inside the callback?)

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -197,34 +197,43 @@ public void Start()
...
+                }
                 try
                 {
-                    _client.ConnectAsync(_logic.UrlRequest.Host, _logic.UrlRequest.Port).ContinueWith(t =>
+                    using (cts)
                     {
```
borrrden: I don't think `using` can be used here because the token source needs to remain valid asynchronously during the connect timeout period.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -197,34 +197,43 @@ public void Start()
...
-                        }
+                        _client.ConnectAsync(_logic.UrlRequest.Host, _logic.UrlRequest.Port).ContinueWith(t =>
+                        {
+                            if (!NetworkTaskSuccessful(t))
+                            {
+                                return;
+                            }
 
-                        _queue.DispatchAsync(StartInternal);
+                            _queue.DispatchAsync(StartInternal);
```
borrrden: Make sure your Visual Studio settings are using spaces, not tabs (for consistency and because it shows up weird like this)

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -197,34 +197,43 @@ public void Start()
...
+                }
+                catch (OperationCanceledException)
+                {
+                    // connection timeout
+                    _client.Close();
                 }
-                catch (Exception e) {
+                catch (SocketException)
+                {
+                    _client.Close();
```
borrrden: I'm not sure what the meaning of this part is.  Did you find some information indicating that the client would not be already closed when a `SocketException` happens?  Also `DidClose` is needed to communicate the closure to LiteCore.

Issue #1012, opened 2018-03-31
Reusing the same collation in a query will cause all of them to be affected by the last modification


borrrden:
Running the tunes perf test revealed this as the following query:

```c#
            var collation = Collation.Unicode().Locale("en_JP").IgnoreCase(true).IgnoreAccents(true);
            QueryBuilder.Select(SelectResult.Expression(Expression.Property("Album")))
                .From(DataSource.Database(Db))
                   .Where(Expression.Property("Artist").Collate(collation).EqualTo(Expression.Parameter("ARTIST"))
                    .And(Expression.Property("Compilation").IsNullOrMissing()))
                .GroupBy(Expression.Property("Album").Collate(collation))
                   .OrderBy(Ordering.Expression(Expression.Property("Album").Collate(collation)))
```

results in the where clause searching for "Album" instead of "Artist" since the reuse later overwrites its state by mistake.

**WORKAROUND** Use different collation instances per call to `Collate` (or at least make one for each unique property name)

PR #1034, opened 2018-06-07
#1012  clone new collation in a query fix issue

Sandychuang8:


src/Couchbase.Lite.Shared/Query/QueryExpression.cs
```diff
@@ -146,7 +146,7 @@ public IExpression Between(IExpression expression1, IExpression expression2)
 
         public IExpression Collate(ICollation collation)
         {
-            var col = Misc.TryCast<ICollation, QueryCollation>(collation);
+            var col = Misc.TryCast<ICollation, QueryCollation>(new QueryCollation((QueryCollation)collation));
```
borrrden: Leave `col` as it is.  It's casting and checking for a valid object.  Then create a copy based on `col`

src/Couchbase.Lite.Tests.Shared/TunesPerfTest.cs
```diff
@@ -254,16 +254,14 @@ private int QueryAlbums(Benchmark bench)
         {
             var albumCount = 0;
             // Workaround for https://github.com/couchbase/couchbase-lite-net/issues/1012
```
borrrden: This comment can be removed

PR #1074, opened 2018-10-26
push filter & pull validator

Sandychuang8:
Adding push filter and pull validator

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -262,7 +262,7 @@ namespace Couchbase.Lite.Sync
             });
         }
 
-        [MonoPInvokeCallback(typeof(C4ReplicatorPushFilterFunction))]
+        [MonoPInvokeCallback(typeof(C4ReplicatorFilterFunction))]
         private static bool PushFilterCallback(FLSlice docID, FLDict* dict, void* context)
         { 
             var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
@@ -270,7 +270,7 @@ namespace Couchbase.Lite.Sync
             return replicator.PushFilterCallback(docIDStr, dict);
         }
 
-        bool PushFilterCallback(string docID, FLDict* dict)
+        private bool PushFilterCallback(string docID, FLDict* dict)
         {
             if (Config.PushFilter == null)
                 return true;
@@ -280,7 +280,7 @@ namespace Couchbase.Lite.Sync
             return v;
         }
 
-        [MonoPInvokeCallback(typeof(C4ReplicatorValidationFunction))]
+        [MonoPInvokeCallback(typeof(C4ReplicatorFilterFunction))]
         private static bool PullValidateCallback(FLSlice docID, FLDict* dict, void* context)
         {
             var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
@@ -288,7 +288,7 @@ namespace Couchbase.Lite.Sync
             return replicator.PullValidateCallback(docIDStr, dict);
         }
 
-        bool PullValidateCallback(string docID, FLDict* dict)
+        private bool PullValidateCallback(string docID, FLDict* dict)
         {
             if (Config.PullFilter == null)
                 return true;
@@ -476,7 +476,7 @@ namespace Couchbase.Lite.Sync
             if(Config.PushFilter!=null)
                 _nativeParams.PushFilter = PushFilterCallback;
             if (Config.PullFilter != null)
-                _nativeParams.Validation = PullValidateCallback;
+                _nativeParams.PullFilter = PullValidateCallback;
             var err = new C4Error();
             var status = default(C4ReplicatorStatus);
             _stopping = false;
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -36,6 +36,8 @@ using JetBrains.Annotations;
 using LiteCore;
 using LiteCore.Interop;
 using LiteCore.Util;
+using Couchbase.Lite.Internal.Serialization;
+using System.Collections.Generic;
 
 namespace Couchbase.Lite.Sync
 {
@@ -270,13 +272,11 @@ namespace Couchbase.Lite.Sync
             return replicator.PushFilterCallback(docIDStr, dict);
         }
 
-        private bool PushFilterCallback(string docID, FLDict* dict)
+        private bool PushFilterCallback(string docID, FLDict* value)
         {
-            if (Config.PushFilter == null)
-                return true;
-            var doc = Config.Database.GetDocument(docID);
+            var d = FLValueConverter.ToCouchbaseObject((FLValue*)value, Config.Database, true) as IDictionary<string, object>;
             var f = Config.PushFilter;
-            var v = f(doc);
+            var v = f(new MutableDocument(docID, d));
             return v;
         }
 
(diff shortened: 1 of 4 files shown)


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -262,10 +262,40 @@ private static void StatusChangedCallback(C4Replicator* repl, C4ReplicatorStatus
...
 
+        [MonoPInvokeCallback(typeof(C4ReplicatorPushFilterFunction))]
+        private static bool PushFilterCallback(FLSlice docID, FLDict* dict, void* context)
+        { 
+            var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
+            var docIDStr = docID.CreateString();
+            return replicator.PushFilterCallback(docIDStr, dict);
+        }
+
+        bool PushFilterCallback(string docID, FLDict* dict)
```
borrrden: Add "private" explicitly

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -262,10 +262,40 @@ private static void StatusChangedCallback(C4Replicator* repl, C4ReplicatorStatus
...
+            var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
+            var docIDStr = docID.CreateString();
+            return replicator.PullValidateCallback(docIDStr, dict);
+        }
+
+        bool PullValidateCallback(string docID, FLDict* dict)
+        {
+            if (Config.PullFilter == null)
+                return true;
+            var doc = Config.Database.GetDocument(docID);
```
borrrden: I'm not sure this is the way that was intended.  What is coming with "dict"?  If it is the document properties then those should be used instead.  Otherwise there will be a database call for every item in the pull replication which will slow things down a lot (ditto for the push callback).

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator.cs
```diff
@@ -32,35 +32,38 @@ namespace LiteCore.Interop
...
+    [return: MarshalAs(UnmanagedType.U1)]
+    internal unsafe delegate bool C4ReplicatorPushFilterFunction(FLSlice docID, FLDict* body,
+        void* context);
 
     [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-    internal unsafe delegate void C4ReplicatorPushFilterFunction(FLSlice docID,
-        FLDict* body, void* context);
+    [return: MarshalAs(UnmanagedType.U1)]
+    internal unsafe delegate bool C4ReplicatorValidationFunction(FLSlice docID, FLDict* body, 
+        void* context);
```
borrrden: These probably don't require different delegates since the signatures are the same.  They are both "Validation functions."

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator.cs
```diff
@@ -61,9 +56,9 @@ internal sealed class ReplicatorParameters : IDisposable
         private C4ReplicatorParameters _c4Params;
         private C4ReplicatorBlobProgressCallback _onBlobProgressUpdated;
         private C4ReplicatorStatusChangedCallback _onStatusChanged;
-        private C4ReplicatorDocumentEndedCallback _onDocumentEnded;
-        private C4ReplicatorPushFilterFunction _pushFilter;
-        private C4ReplicatorValidationFunction _validation;
+        private C4ReplicatorDocumentErrorCallback _onDocumentEnded;
```
borrrden: Why does this keep changing back and forth?

Sandychuang8: I modified C4ReplicatorDocumentErrorCallback to C4ReplicatorDocumentEndedCallback in feature replication eventing. But this PR should not have that change yet...so I changed it back..

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: I actually don't know how to get info out FLDict.

borrrden: You could use FLValueConverter to convert it to a .NET dictionary and examine in the debugger.  Or ask Jens / Pasin what this dict is supposed to contain.

Sandychuang8: remove Config.PullFilter == null because it will never get to here is Config.PullFilter == null

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -116,6 +117,26 @@ public void TestEmptyPush()
             RunReplication(config, 0, 0);
         }
 
+        [Fact]
+        public void TestPushDocWithFilter()
+        {
+            using (var doc1 = new MutableDocument("doc1"))
```
borrrden: A better test would be to set up a database with two documents, and set up a filter that rejects one of them.  Then you can check that only one document exists in the target database (`_otherDB`).

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -161,6 +183,27 @@ public void TestPushDocContinuous()
             }
         }
 
+        [Fact]
+        public void TestPullDocWithFilter()
+        {
+            using (var doc1 = new MutableDocument("doc1")) {
```
borrrden: Same as the pushdocwithfilter comment for this one, except the reverse (insert into `_otherDB` and check `Db`)

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
Sandychuang8: you are right. updated.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
Sandychuang8: updated.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -905,6 +950,17 @@ private ReplicatorConfiguration CreateConfig(bool push, bool pull, bool continuo
         }
 #endif
 
+        private bool _replicator__filterCallback(Document doc)
+        {
+            _isFilteredCallback = true;
+            var name = doc.GetString("name");
+            if (name == "pass") {
```
borrrden: `return name == "pass"` ?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -116,6 +117,27 @@ public void TestEmptyPush()
...
+                doc2.SetString("name", "pass");
+                Db.Save(doc2);
+            }
+
+            var config = CreateConfig(true, false, false);
+            config.PushFilter = _replicator__filterCallback;
+            RunReplication(config, 0, 0);
+            _isFilteredCallback.Should().BeTrue();
+            _otherDB.GetDocument("doc1").Should().BeNull("because doc1 is filtered out in the callback");
+            _otherDB.GetDocument("doc2").Count().Should().Be(1, "because doc2 is filtered in in the callback");
```
borrrden: This logic is slightly hard to follow.  For these purposes it is enough if the document simply is not null (other tests will test that the push operation actually transmits data correctly)

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -161,6 +184,27 @@ public void TestPushDocContinuous()
...
+                doc2.SetString("name", "pass");
+                _otherDB.Save(doc2);
+            }
+
+            var config = CreateConfig(false, true, false);
+            config.PullFilter = _replicator__filterCallback;
+            RunReplication(config, 0, 0);
+            _isFilteredCallback.Should().BeTrue();
+            Db.GetDocument("doc1").Should().BeNull("because doc1 is filtered out in the callback");
+            Db.GetDocument("doc2").Count().Should().Be(1, "because doc2 is filtered in in the callback");
```
borrrden: Same comment as push test.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -69,6 +69,8 @@ public sealed class ReplicatorConfiguration
         private Authenticator _authenticator;
         private bool _continuous;
         private ReplicatorType _replicatorType = ReplicatorType.PushAndPull;
+        private Func<Document, bool> _PushFilter;
+        private Func<Document, bool> _PullValidator;
```
borrrden: underscore lower case for consistency

PR #1076, opened 2018-10-30
adding option to disable the Delta Syncs

Sandychuang8:
adding option to disable the Delta Syncs


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -444,6 +449,7 @@ private void StartInternal()
             // Clear the reset flag, it is a one-time thing
             Config.Options.Reset = false;
 
+            Config.Options.NoDeltas = false;
```
borrrden: Why is this here?  The reset flag should be set only once because it's not something that should happen every time but this is presumably a setting that we want to keep on right?

Issue #1083, opened 2018-11-20
Purge Document with ID and Set Expiration should allow to operate on deleted documents


pasin:
### Library Version

Iridium

### .NET Runtime

Any

### Operating System / Device Details

Any

### Expected behavior

Documents including deleted documents should allow to be purged or set expiration.

### Actual behavior

Based on [this](https://github.com/couchbase/couchbase-lite-net/blob/master/src/Couchbase.Lite.Shared/API/Database/Database.cs#L748) and [this](https://github.com/couchbase/couchbase-lite-net/blob/master/src/Couchbase.Lite.Shared/API/Database/Database.cs#L765-L767), I understand that purging or setting expiration on deleted documents may not be working as `GetDocument(doc)` will return `null` for deleted documents?

Issue #1082, opened 2018-11-20
2.x TTL doesn't purge expired documents


borrrden:
### Library Version

Iridium

### .NET Runtime

All

### Operating System / Device Details

All

### Log Output

n/a

### Expected behavior

An expired document should be purged

### Actual behavior

No purge occurs

### Steps To Reproduce

1. Set TTL to something small (e.g. 2 seconds)
2. Wait for more than the TTL
3. Observe that the document is still present

borrrden:
@Sandychuang8 I blanked on this when reviewing the PR.  When document expiration is set, a timer needs to be set up for the next expiration (unless a timer is already set for a point in time sooner than the new expiration).  The timer's job is to then run the LiteCore `c4db_purgeExpiredDocs` method, and reschedule if there is another document that will be expiring.  But be careful not to exceed the maximum allowed time for a timer (see https://github.com/couchbase/couchbase-lite-net/issues/1029 )

PR #1084, opened 2018-11-21
#1082 #1083 purge documents when docs expiration is set and reached, allow deleted doc to be purged

Sandychuang8:


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -745,8 +749,13 @@ public void Purge(Document document)
         public void Purge(string docId)
         {
             CBDebug.MustNotBeNull(Log.To.Database, Tag, nameof(docId), docId);
-            var document = GetDocument(docId);
-            Purge(document);
+            try {
+                LiteCoreBridge.Check(err => Native.c4db_beginTransaction(_c4db, err));
+                PurgeDocById(docId);
+                LiteCoreBridge.Check(e => Native.c4db_endTransaction(_c4db, true, e));
+            } catch (Exception) {
```
borrrden: This is not the best way to do this.  Use try / finally like other places and move the end_transaction to "finally" so that the exception is thrown up to the caller.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -951,6 +958,29 @@ internal void ResolveConflict([NotNull]string docID)
                 }
             });
         }
+
+        internal void SchedulePurgeExpired(TimeSpan delay)
+        {
+            var nextExpiration = Native.c4db_nextDocExpiration(_c4db);
+            if (nextExpiration > 0) {
+                var delta = (DateTimeOffset.FromUnixTimeMilliseconds((long)nextExpiration) - DateTimeOffset.UtcNow).Add(TimeSpan.FromSeconds(1));
```
borrrden: What is the reason for adding one second?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -951,6 +958,29 @@ internal void ResolveConflict([NotNull]string docID)
...
+                var expirationTimeSpan = delta > delay ? delta : delay;
+                if (expirationTimeSpan.TotalMilliseconds >= UInt32.MaxValue) {
+                    _expirePurgeTimer.Change(TimeSpan.FromMilliseconds(UInt32.MaxValue - 1), TimeSpan.FromMilliseconds(-1));
+                    Log.To.Database.I(Tag, "{0:F3} seconds is too far in the future to schedule a document expiration," +
+                                           " will run again at the maximum value of {0:F3} seconds", expirationTimeSpan.TotalSeconds, (UInt32.MaxValue - 1) / 1000);
+                } else if (expirationTimeSpan.TotalSeconds <= Double.Epsilon) {
+                    _expirePurgeTimer.Change(Timeout.Infinite, Timeout.Infinite);
+                    PurgeExpired(null);
+                } else {
+                    _expirePurgeTimer.Change(expirationTimeSpan, TimeSpan.FromMilliseconds(-1));
```
borrrden: You should probably check what the existing timer is set to as well, to avoid needless calculation.  If the timer is already going to run sooner than the new delay then there is no need to reschedule.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1365,6 +1397,29 @@ private void VerifyDB([NotNull]Document document)
...
+        {
+            var cnt = 0L;
+            ThreadSafety.DoLockedBridge(err =>
+            {
+                CheckOpen();
+                cnt = Native.c4db_purgeExpiredDocs(_c4db, err);
+                Log.To.Database.I(Tag, "{0} purged {1} expired documents", this, cnt);
+                SchedulePurgeExpired(TimeSpan.FromSeconds(1));
+                if (err->code>0 && err->domain>0)
+                    return false;
```
borrrden: Where is this return value going and why is it `false` for success?

borrrden:
Also this PR does a lot more than "add purge document" so please add more details to messages in the future.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: it's false when error code and domain are greater than 0

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -749,13 +749,18 @@ public void Purge(Document document)
...
-            } catch (Exception) {
-                LiteCoreBridge.Check(e => Native.c4db_endTransaction(_c4db, false, e));
-            }
+                try {
+                    PurgeDocById(docId);
+                } catch (Exception e) {
+                    Log.To.Database.W(Tag, "Exception during Purge doc by id...", e);
+                    throw;
+                } finally {
+                    LiteCoreBridge.Check(e => Native.c4db_endTransaction(_c4db, false, e));
```
borrrden: New problem:  This will never save the changes since end transaction always uses false.  I'd say just call `InBatch` instead of `ThreadSafety.DoLocked` since it is designed to do the same thing.  It both locks and runs an item inside a transaction `InBatch(() => PurgeDocById(docId));`

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: Oh, I saw it as `=` somehow, but regardless this return value is unused.  Just remove the logic for returning true or false and call with `LiteCoreBridge.Check` instead

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: because of SchedulePurgeExpired(TimeSpan.FromSeconds(1));

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -749,13 +749,7 @@ public void Purge(Document document)
...
         {
             CBDebug.MustNotBeNull(Log.To.Database, Tag, nameof(docId), docId);
-            try {
-                LiteCoreBridge.Check(err => Native.c4db_beginTransaction(_c4db, err));
-                PurgeDocById(docId);
-                LiteCoreBridge.Check(e => Native.c4db_endTransaction(_c4db, true, e));
-            } catch (Exception) {
-                LiteCoreBridge.Check(e => Native.c4db_endTransaction(_c4db, false, e));
-            }
+            ThreadSafety.DoLocked(() => InBatch(() => PurgeDocById(docId)));
```
borrrden: `InBatch` already does `ThreadSafety.DoLocked`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -797,9 +791,8 @@ public bool SetDocumentExpiration(string docId, DateTimeOffset? timestamp)
         /// doesn't exist</exception>
         public DateTimeOffset? GetDocumentExpiration(string docId)
         {
-            if (GetDocument(docId) == null) {
-                throw new CouchbaseLiteException(C4ErrorCode.NotFound, "Cannot find the document.");
-            }
+            if (LiteCoreBridge.Check(err => Native.c4doc_get(_c4db, docId, true, err)) == null)
+                throw new CouchbaseLiteException(C4ErrorCode.NotFound);
```
borrrden: Don't omit brackets for one line if statements.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1408,16 +1401,14 @@ private void PurgeDocById(string id)
...
+            LiteCoreBridge.Check(err =>
             {
                 CheckOpen();
                 cnt = Native.c4db_purgeExpiredDocs(_c4db, err);
                 Log.To.Database.I(Tag, "{0} purged {1} expired documents", this, cnt);
-                SchedulePurgeExpired(TimeSpan.FromSeconds(1));
-                if (err->code>0 && err->domain>0)
-                    return false;
-                return true;
+                return err;
```
borrrden: Why are you returning err?  You don't need to return anything.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: needs 1 sec increment or it will stuck in a loop

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: it complained if I don't return anything.

borrrden: Ok based on offline discussion it should be err.code == 0

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -957,7 +957,7 @@ internal void SchedulePurgeExpired(TimeSpan delay)
         {
             var nextExpiration = Native.c4db_nextDocExpiration(_c4db);
             if (nextExpiration > 0) {
-                var delta = (DateTimeOffset.FromUnixTimeMilliseconds((long)nextExpiration) - DateTimeOffset.UtcNow).Add(TimeSpan.FromSeconds(1));
+                var delta = (DateTimeOffset.FromUnixTimeMilliseconds((long)nextExpiration) - DateTimeOffset.UtcNow).Add(TimeSpan.FromSeconds(0));
```
borrrden: Adding zero is redundant, just remove the `Add`.

borrrden:
The API has changed slightly.  Could you review the latest commit and modify this PR?

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
```diff
@@ -1852,6 +1921,24 @@ public void TestGetExpirationFromNoneExistDoc()
...
+        public void TestLongExpiration()
+        {
+            var now = DateTime.UtcNow;
+            using (var doc = new MutableDocument("doc")) {
+                doc.SetInt("answer", 42);
+                doc.SetValue("options", new[] { 1, 2, 3 });
+                Db.Save(doc);
+
+                Db.GetDocumentExpiration("doc").Should().BeNull();
+                Db.SetDocumentExpiration("doc", new DateTimeOffset(DateTime.UtcNow.AddDays(60)));
```
borrrden: This is an odd way to do it.  Why not just `DateTimeOffset.UtcNow.AddDays`?

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: how to get current expiration timer's total milliseconds from _expirePurgeTimer?

borrrden: I thought there was a way but I was wrong, so this thread is not irrelevant.  Just double check on the `else if` condition here. Should it be milliseconds instead of seconds?

PR #1089, opened 2018-12-05
API update for replication eventing

Sandychuang8:
Replication eventing: update the document ended listener name to AddDocumentReplicationListener and use RemoveChangeListener to remove both/either ChangeListener & DocumentReplicationListener

Replication filters:  update LiteCore, update C4ReplicatorFilterFunction name to C4ReplicatorValidationFunction, add ReplicationFilter class which contains Document and IsDeleted properties, add C4RevisionFlags parameter in filter callback function, add push replication with deleted document test

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -135,11 +135,11 @@ namespace Couchbase.Lite.Sync
         /// <param name="handler">The logic to run during the callback</param>
         /// <returns>A token to remove the handler later</returns>
         [ContractAnnotation("null => halt")]
-        public ListenerToken AddReplicationListener(EventHandler<DocumentReplicatedEventArgs> handler)
+        public ListenerToken AddDocumentReplicationListener(EventHandler<DocumentReplicatedEventArgs> handler)
         {
             CBDebug.MustNotBeNull(Log.To.Sync, Tag, nameof(handler), handler);
 
-            return AddReplicationListener(null, handler);
+            return AddDocumentReplicationListener(null, handler);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -206,12 +196,16 @@ public ListenerToken AddChangeListener(EventHandler<ReplicatorStatusChangedEvent
...
-        /// Removes a previously added change listener via its <see cref="ListenerToken"/>
+        /// Removes a previously added change listener via its <see cref="ListenerToken"/> and/or
+        /// Removes a previously added documents ended listener via its <see cref="ListenerToken"/>
         /// </summary>
-        /// <param name="token">The token received from <see cref="AddChangeListener(TaskScheduler, EventHandler{ReplicatorStatusChangedEventArgs})"/></param>
+        /// <param name="token">The token received from <see cref="AddChangeListener(TaskScheduler, EventHandler{ReplicatorStatusChangedEventArgs})"/>
+        /// and/or The token received from <see cref="AddReplicationListener(TaskScheduler, EventHandler{DocumentReplicatedEventArgs})"/></param>
         public void RemoveChangeListener(ListenerToken token)
         {
             _statusChanged.Remove(token);
```
borrrden: I'm not sure this will be ok with the wrong token type.  You should check the type and perform the correct removal.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -74,6 +74,12 @@ internal enum ReplicatorProgressLevel : int
         PerAttachment // >=2
     }
 
+    public class ReplicationFilter
```
borrrden: This is awkward and something I argued against when designing the API, so in cases like this we are allowed on .NET to just use functions (which is why there is a comment there saying `/* Use block or function instead if possible */`).  Revert this back to the way it was

Sandychuang8: how do I pass isDeleted flag in the callback if use the way it was?

borrrden: It should be `Func<bool, Document, bool>` so that it is a function that takes a `Document` and a `bool`

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: It works as is. The token passed into the Remove method, it returns 0 if no token with the correct id to remove or if the token with the correct id is removed. If not the correct id Token pass into the method and token exists, Remove method will return 1.

borrrden: Ok, I was concerned about the token being `null` inside but if you have tested that this combined method works then it should be fine.

PR #1095, opened 2018-12-10
#1090 TTL purge timer issue: stop timer, dispose timer, and set timer…

Sandychuang8:
#1090 TTL purge timer issue: 
stop timer, dispose timer, and set timer object to null. Forcefully stop the database after set document expiration complete to see if test will still pass. Before the fix, forcefully stop the database after set document expiration complete will fail the test. After the fix, the test is passed.


borrrden:
Did you run tests on this PR?  It's important to run many tests so that the test runner is running long enough to reproduce the TTL issue.

Issue #1100, opened 2018-12-19
Replication Eventing Status ToString() method is not returning complete object in string format


hrajput89:
### Library Version
2.5.0-b0091

### .NET Runtime
.Net Core 2.5.0-b0091

### Operating System / Device Details
Windows 10

### Log Output
NA

### Expected behavior

change.Status.ToString() should return all the attributes of the object in string format
E.g. 

> DocumentReplication {replicator=Replicator[-> Database@7960e21{name='cbl-test1545202672.26'} URLEndpoint{url=ws://192.168.0.126:4984/db}], is pushing =true, document id =push_cbl_docs_3, error code =0, error domain=0, doc is deleted =false}

### Actual behavior
"Couchbase.Lite.Sync.DocumentReplication"

<img width="1440" alt="screenshot 2018-12-19 13 41 59" src="https://user-images.githubusercontent.com/18100038/50207241-e6df9900-0393-11e9-9b51-4223b857ccdd.png">

Steps To Reproduce

Create docs in CBL, Create docs in SGW
2. Add Document Replication Change listener for replicator.
3. Replicate docs push/pull to SGW through one shot replication
4. Get the events of replicator from listener
5. Verify the event matches with the expected outcome.
i. Assert the total event count should be equal to no. of docs replicated
ii. Assert no. push event are equal to pushed docs
iii. Assert no. push event are equal to pushed docs
iv. Assert there is no error in events
Reproduction Project

https://github.com/couchbaselabs/mobile-testkit/tree/feature/cbl2.5-replication-eventing

pytest -rsx -s --no-conflict --timeout 3600 --liteserv-version=2.5.0-91 --liteserv-host=172.23.121.57 --liteserv-port=8080 --sync-gateway-version=2.5.0-132 --mode=cc --server-version=6.5.0-1813 --liteserv-platform=net-msft --create-db-per-test=cbl-test testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication_eventing.py -k test_replication_eventing_status

borrrden:
This is not a functional requirement, but rather something that you wish to leverage for the mobile testkit app.

PR #1104, opened 2019-01-03
Override ToString method in ReplicatedDocument class.

Sandychuang8:
#1100 Override ToString method in ReplicatedDocument class. This will help QA to print out the debug info easily.


Sandychuang8:
close #1100

PR #1117, opened 2019-01-18
Feature/persist data type of blob in predictive query

Sandychuang8:
#1111 Blob metadata not preserved in predictive query parameters fixes.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Document/Blob.cs
+++ b/src/Couchbase.Lite.Shared/API/Document/Blob.cs
@@ -135,13 +135,10 @@ namespace Couchbase.Lite
                 if(_db != null) {
                     C4BlobStore* blobStore;
                     C4BlobKey key;
-                    if(!GetBlobStore(&blobStore, &key)) {
-                        return null;
+                    if(GetBlobStore(&blobStore, &key)) {
+                        return new BlobReadStream(blobStore, key);
                     }
-
-                    return new BlobReadStream(blobStore, key);
                 }
-
                 return _content != null ? new MemoryStream(_content) : null;
             }
         }
--- a/src/Couchbase.Lite.Shared/API/Document/Blob.cs
+++ b/src/Couchbase.Lite.Shared/API/Document/Blob.cs
@@ -46,6 +46,7 @@ namespace Couchbase.Lite
         private const string ContentTypeKey = "content_type";
         private const string DigestKey = "digest";
         private const string LengthKey = "length";
+        private const string DataKey = "data";
 
         private const uint MaxCachedContentLength = 8 * 1024;
         private const int ReadBufferSize = 8 * 1024;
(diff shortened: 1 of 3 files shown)


src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -257,9 +257,10 @@ internal Blob([NotNull]Database db, [NotNull]IDictionary<string, object> propert
             _db = CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(db), db);
             _properties = new Dictionary<string, object>(CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, 
                 nameof(properties), properties));
+            _content = properties.GetCast<byte[]>(DataKey);
             ContentType = properties.GetCast<string>(ContentTypeKey);
-            if(Digest == null) {
-                WriteLog.To.Database.W(Tag, "Blob read from database has missing digest");
+            if(Digest == null && _content == null) {
+                WriteLog.To.Database.W(Tag, "Blob read from database has missing digest and contents.");
```
borrrden: Wording issue.  "has neither digest nor data" would be better

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -280,12 +281,9 @@ internal void FLEncode(FLEncoder* enc)
                     WriteLog.To.Database.W(Tag, "Error installing blob to database, throwing...");
                     throw;
                 }
-
-                var dict = JsonRepresentation;
-                dict.FLEncode(enc);
+                JsonRepresentation.FLEncode(enc);
```
borrrden: Since this is used directly in both cases, just move it outside the `if`

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -280,12 +281,9 @@ internal void FLEncode(FLEncoder* enc)
...
             } else {
-                // Independent blob, write bytes only
-                ((IEnumerable<byte>)Content).FLEncode(enc);
+                JsonRepresentation.FLEncode(enc);
```
borrrden: And remove the `else` as well

src/Couchbase.Lite.Shared/Serialization/FLValueConverter.cs
```diff
@@ -190,7 +190,9 @@ private static object ToObject(FLValue* value, Database db, int level = 0, Type
                         retVal[key] = ToObject(Native.FLDictIterator_GetValue(&i), db, level + 1, hintType1);
                     } while (Native.FLDictIterator_Next(&i));
 
-                    return ConvertDictionary(retVal as IDictionary<string, object>, db) ?? retVal;
+                    return (string)retVal["@type"] == "blob"
```
borrrden: We have constants set up for this kind of stuff ("@type" and "blob") but they are not needed here.  `ConvertDictionary` already checks this.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
Sandychuang8: Right, duh!

PR #1138, opened 2019-04-08
Fix getting column title string in query

Sandychuang8:


borrrden:
Are all of the tests passing?  Appveyor will fail regardless but it looks like at least one test is failing still on the branch (TestQueryProjectingKeys)

borrrden:
I checked out and ran this branch but tests are still crashing.  TestJoinByDocID, for one.  Please verify that the entire set of tests passes.

Sandychuang8:
All tests are passed now.

src/Couchbase.Lite.Tests.Shared/QueryTest.cs
```diff
@@ -1020,7 +1020,7 @@ public void TestQueryProjectingKeys()
                     r.GetDouble("$1").Should().Be(r.GetDouble(0));
                     r.GetInt("$2").Should().Be(r.GetInt(1));
                     r.GetInt("min").Should().Be(r.GetInt(2));
-                    r.GetInt("$3").Should().Be(r.GetInt(3));
+                    r.GetInt("$4").Should().Be(r.GetInt(3));
```
borrrden: This doesn't seem like the correct result.  This needs to be verified.

PR #1139, opened 2019-04-16
Update Query Column Title

Sandychuang8:


borrrden:
Also let's add another test if it is not present to try using `QueryBuilder.Select().From(...)` (the select clause is empty, which is an implicit select *)

src/Couchbase.Lite.Tests.Shared/QueryTest.cs
```diff
@@ -2342,6 +2342,21 @@ public void TestMillisConversion()
...
+            int resultCnt = 0;
+            string k1 = "";
+            string k2 = "";
+            LoadNumbers(100);
+            using (var q = QueryBuilder.Select().From(DataSource.Database(Db))
+                .Where(Expression.Property("number1").LessThan(Expression.Int(10))).OrderBy(Ordering.Property("number1"))) {
+                var res = q.Execute();
+                resultCnt = res.Count();
+            }
+            resultCnt.Should().Be(9);
```
borrrden: The test is for the naming of the column so please check the name of the column and make sure it is correct.

src/Couchbase.Lite.Shared/Query/XQuery.cs
```diff
@@ -207,13 +207,12 @@ private unsafe void Check()
...
-                    if(selectListCnt == columnCnt)
-                        titleStr = selectResultList.ElementAtOrDefault(i)?.ColumnName;
-                }
 
-                if (String.IsNullOrEmpty(titleStr))
-                    titleStr = FromImpl.ColumnName;
+                if (titleStr.StartsWith("*")) {
+                    titleStr = selectResultList.ElementAtOrDefault(i)?.ColumnName;
+                    if (String.IsNullOrEmpty(titleStr))
+                        titleStr = FromImpl.ColumnName;
```
borrrden: Double check that this is needed....I don't think it is

Sandychuang8: Actually it was hit several times...

borrrden: Interesting.  Was there any times that line 211 was true but line 213 was false?  Could we remove the "column name" completely?

Sandychuang8: Yes, few times that line 211 was true but line 213 was false. Cannot remove "column name" completely.

borrrden: I'm not getting why we need to have both.....It seems overly complicated.  Can you give examples of each case?

PR #1151, opened 2019-05-09
Feature/conflict resolver

Sandychuang8:


src/Couchbase.Lite.Shared/API/Conflict/IConflictResolver.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Couchbase.Lite
+{
+    /// <summary>
+    /// Conflict Resolver Interface
+    /// </summary>
+    public interface IConflictResolver
+    {
+        /// <summary>
+        /// The callback conflict resolve method, if conflict occurs.
```
borrrden: Mention the limitations and under what circumstances an exception will be thrown in the doc comments

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -871,14 +871,21 @@ internal void ResolveConflict([NotNull]string docID)
                         WriteLog.To.Database.I(Tag, "Resolving doc '{0}' (mine={1} and theirs={2})",
                             new SecureLogString(docID, LogMessageSensitivity.PotentiallyInsecure), localDoc.RevID,
                             remoteDoc.RevID);
-                        var resolvedDoc = ResolveConflict(localDoc, remoteDoc);
+                        conflictResolver = conflictResolver == null ? new DefaultConflictResolver() : conflictResolver;
```
borrrden: `conflictResolver ?? new DefaultConflictResolver()`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -871,14 +871,21 @@ internal void ResolveConflict([NotNull]string docID)
...
+                        var conflict = new Conflict(localDoc.IsDeleted ? null : localDoc, remoteDoc.IsDeleted ? null : remoteDoc);
+                        var resolvedDoc = conflictResolver.Resolve(conflict);
+                        if (resolvedDoc != null && resolvedDoc.Id != docID)
+                            throw new Exception($"Resolved docID {resolvedDoc.Id} does not match with docID {docID}");
```
borrrden: Pick a better exception type then just generic `Exception`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1265,40 +1250,60 @@ private void Save([NotNull]Document doc, C4Document** outDoc, C4Document* baseDo
...
+            if (resolvedDoc == null) {
+                if (localDoc.IsDeleted)
+                    resolvedDoc = localDoc;
+
+                if (remoteDoc.IsDeleted)
+                    resolvedDoc = remoteDoc;
+            }
+
+
+            if (resolvedDoc!=null&&!ReferenceEquals(resolvedDoc, localDoc)) {
```
borrrden: Spacing issues with `&&`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1265,40 +1250,60 @@ private void Save([NotNull]Document doc, C4Document** outDoc, C4Document* baseDo
...
+                        WriteLog.To.Sync.E(Tag, "Resolved document for {0} somehow got disposed!",
+                            new SecureLogString(resolvedDoc.Id, LogMessageSensitivity.PotentiallyInsecure));
+                        throw new RuntimeException(
+                            "Resolved document was disposed before conflict resolution completed.  Please file a bug report at https://github.com/couchbase/couchbase-lite-net");
+                    }
+                    if (mergedBody.Equals((FLSliceResult)FLSlice.Null))
+                        throw new RuntimeException("Resolved document contains a null body");
+                    isDeleted = resolvedDoc.IsDeleted;
+                } else {
+                    mergedBody = (FLSliceResult)FLSlice.Allocate("");
```
borrrden: Is there a reason to use `""`?  If it has to be used instead of `null` then this makes me nervous.  Mixing allocations from C# with frees from LiteCore is not a good idea.

src/Couchbase.Lite.Shared/API/Document/Document.cs
```diff
@@ -185,7 +185,7 @@ internal Document([NotNull]Document other)
         /// can be edited)
         /// </summary>
         /// <returns>A mutable version of the document</returns>
-        [NotNull]
+        [CanBeNull]
```
borrrden: Why?  This is still never null.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1078,16 +1067,162 @@ public void TestExpiredNotPushed()
...
+                return null;
+            });
+
+            RunReplication(config, 0, 0);
+
+            Db.Count.ShouldBeEquivalentTo(0);
+        }
+
+        [Fact]
+        public void TestConflictResolverDeletedLocalWin()
```
borrrden: Not sure what this is testing?

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Since mergedBody initialized as FLSlice.Null, I guess I can remove the logic ```else {
                    mergedBody = (FLSliceResult)FLSlice.Allocate("");
                }``` ?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
Sandychuang8: This is to test the scenario ```if (localDoc.IsDeleted)
                    resolvedDoc = localDoc;```

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: After tested with setting mergedBody to FLSlice.Null, ConflictResolver returns null test failed.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
borrrden: Why is it needed though?  All the resolver does is return null....

src/Couchbase.Lite.Shared/API/Conflict/IConflictResolver.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// When a null document is returned, the conflict will be resolved as document deletion. 
+        /// If there is an exception thrown in the resolve method, the exception will be caught and handled:
+        /// The conflict resolving will be skipped. The pending conflicted documents will be resolved when the replicator is restarted.
+        /// </summary>
+        /// <exception cref="RuntimeException">Thrown when a document with the wrong ID is returned.</exception> 
+        /// <exception cref="RuntimeException">Thrown when a document from a different database is returned.</exception>
```
borrrden: Isn't this `InvalidOperationException` ?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -871,14 +871,21 @@ internal void ResolveConflict([NotNull]string docID)
...
                         SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
                     } finally {
                         localDoc?.Dispose();
                         remoteDoc?.Dispose();
                     }
-                } catch (Exception) {
+                } catch (Exception ex) {
                     success = false;
+                    WriteLog.To.Sync.W(Tag, $"Exception in conflict resolver: {ex.Message}.");
                     throw;
```
borrrden: Didn't the doc comments say this is supposed to be handled?  Do we have a test that throws an exception in the conflict handler?

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: Fails how?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1265,40 +1250,62 @@ private void Save([NotNull]Document doc, C4Document** outDoc, C4Document* baseDo
...
+                    } catch (ObjectDisposedException) {
+                        WriteLog.To.Sync.E(Tag, "Resolved document for {0} somehow got disposed!",
+                            new SecureLogString(resolvedDoc.Id, LogMessageSensitivity.PotentiallyInsecure));
+                        throw new RuntimeException(
+                            "Resolved document was disposed before conflict resolution completed.  Please file a bug report at https://github.com/couchbase/couchbase-lite-net");
+                    }
+                    if (mergedBody.Equals((FLSliceResult)FLSlice.Null))
+                        throw new RuntimeException("Resolved document contains a null body");
+                    isDeleted = resolvedDoc.IsDeleted;
+                } else if (resolvedDoc == null || resolvedDoc.IsEmpty) {
```
borrrden: Looks like this check doesn't make as much sense as I thought.  This is an `else` against `resolvedDoc != null` so it is guaranteed to be `null` here.  I guess just a simple `else` will do.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1265,40 +1250,62 @@ private void Save([NotNull]Document doc, C4Document** outDoc, C4Document* baseDo
...
+                    Native.FLEncoder_Reset(encoder);
                 }
+
+                if (isDeleted)
+                    mergedFlags |= C4RevisionFlags.Deleted;
             }
 
             // Tell LiteCore to do the resolution:
             C4Document* rawDoc = localDoc.c4Doc != null ? localDoc.c4Doc.RawDoc : null;
-            var flags = resolved.c4Doc != null ? resolved.c4Doc.RawDoc->selectedRev.flags : 0;
```
borrrden: This part worries me a bit because any other flags that happen to be in `resolved` are thrown away.  I think `mergedFlags` should be initialized to this line instead of just initialized to 0 above.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -231,6 +232,12 @@ public ReplicatorType ReplicatorType
         [NotNull]
         public IEndpoint Target { get; }
 
+        public IConflictResolver ConflictResolver
```
borrrden: `[CanBeNull]`

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: No tests for testing exception yet. Will add them with other advanced test cases in other PR.

borrrden: But without them there is no way of knowing if this is correct, which it seems like it is not.

src/Couchbase.Lite.Shared/API/Database/Database.cs
jayahariv: this happens, when the local copy is deleted and resolvedDoc will be null and if we try to save a null body through `c4doc_resolveConflict `, it will end up [here](https://github.com/couchbase/couchbase-lite-core/blob/ecf16ecb29ac139e7af8e7d411ac8631344cdd91/LiteCore/Database/TreeDocument.cc#L304) without entering inside it, to save the new revision.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -871,14 +871,21 @@ internal void ResolveConflict([NotNull]string docID)
                         WriteLog.To.Database.I(Tag, "Resolving doc '{0}' (mine={1} and theirs={2})",
                             new SecureLogString(docID, LogMessageSensitivity.PotentiallyInsecure), localDoc.RevID,
                             remoteDoc.RevID);
-                        var resolvedDoc = ResolveConflict(localDoc, remoteDoc);
+                        conflictResolver = conflictResolver ?? new DefaultConflictResolver();
+                        var conflict = new Conflict(localDoc.IsDeleted ? null : localDoc, remoteDoc.IsDeleted ? null : remoteDoc);
+                        var resolvedDoc = conflictResolver.Resolve(conflict);
+                        if (resolvedDoc != null && resolvedDoc.Id != docID)
+                            throw new InvalidOperationException($"Resolved docID {resolvedDoc.Id} does not match with docID {docID}");
+                        if (resolvedDoc != null && resolvedDoc.Database != this)
```
jayahariv: just to note, if the resolve function returns a newly created document, then the resolvedDoc.database might be null. which might cause this to throw.

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: This log statement appears unneeded since it is logged again further up.  One of them should be removed (the other is in Replication.cs)

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -850,47 +850,50 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
...
             {
-                var success = true;
-                LiteCoreBridge.Check(err => Native.c4db_beginTransaction(_c4db, err));
-                Document localDoc = null, remoteDoc = null;
-                try {
-                    localDoc = new Document(this, docID);
-                    if (!localDoc.Exists) {
-                        throw new CouchbaseLiteException(C4ErrorCode.NotFound);
-                    }
+                while (true) {
```
borrrden: This looks like an infinite loop?  When will it exit besides with an exception?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1300,21 +1300,17 @@ private bool SaveResolvedDocument([CanBeNull]Document resolvedDoc, [NotNull]Docu
...
-                            losingRevID_.AsFLSlice(), (FLSlice)mergedBody, mergedFlags, err);
-                        Native.FLSliceResult_Release(mergedBody);
-                        return retVal;
-                    });
+                C4Error err;
+                var retVal = NativeRaw.c4doc_resolveConflict(rawDoc, winningRevID_.AsFLSlice(),
+                    losingRevID_.AsFLSlice(), (FLSlice)mergedBody, mergedFlags, &err);
+                Native.FLSliceResult_Release(mergedBody);
+                
+                if(!retVal && err.code == (int)C4ErrorCode.Conflict)
```
borrrden: If the err.code is not conflict throw an exception

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
borrrden: Make sure that the test is checking if the "deleted" document passed to the resolution function is `null` as it should be.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -847,34 +847,31 @@ public void Save(Couchbase.Lite.Linq.IDocumentModel model)
         internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolver conflictResolver)
         {
             Debug.Assert(docID != null);
+            var saved = false;
```
borrrden: This variable doesn't serve a purpose I feel.  It should be simply ok to move `success` here outside the loop and use it directly.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -884,18 +884,17 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
                         InBatch(() =>
                         {
                             success = SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
-
                         });
                         if (!success)
                             continue;
```
borrrden: This continue does not serve any purpose.

PR #1152, opened 2019-05-11
Add conflict handler for saving document with conflicts into database…

Sandychuang8:
… and a simple unit test to test feature

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -576,7 +576,7 @@ namespace Couchbase.Lite
         public bool Delete([NotNull]Document document, ConcurrencyControl concurrencyControl)
         {
             var doc = CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(document), document);
-            return Save(doc, concurrencyControl, true);
+            return Save(doc, null, concurrencyControl, true);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -816,10 +816,79 @@ public void RemoveChangeListener(ListenerToken token)
...
+            Debug.Assert(conflictHandler != null);
+            Document currDoc = null;
+
+            while (true) {
+                var success = Save(doc, currDoc, ConcurrencyControl.FailOnConflict, false);
+                if (success)
+                    return success;
+                var committed = false;
+                try {
+                    LiteCoreBridge.Check(err => Native.c4db_beginTransaction(_c4db, err));
```
borrrden: You cannot hold this transaction the whole time since that will block write access to the database during conflict resolution.  The attempted save should be inside of `DoLocked` and a transaction.  Then if the result is false, the conflict resolution needs to take place outside the lock and transaction, and then loop around again into another lock and transaction.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -816,10 +816,55 @@ public void RemoveChangeListener(ListenerToken token)
...
+                    var isDeleted = curDoc->flags.HasFlag(C4DocumentFlags.DocDeleted);
+                    if (isDeleted) {
+                        oldDoc.ReplaceC4Doc(new C4DocumentWrapper(curDoc));
+                        Native.c4doc_free(curDoc);
+                        curDoc = null;
+                    }
+                    if (oldDoc == null) {
+                        return false;
+                    }
+                    currDoc = isDeleted ? null : oldDoc;
```
borrrden: I can't understand what is going on with any of the lines between `if(!saved)` and here...

What is their purpose?

Also `curDoc` will be leaked if `isDeleted` is `false`

src/Couchbase.Lite.Tests.Shared/DatabaseTest.cs
```diff
@@ -305,6 +307,104 @@ public void TestSaveConcurrencyControl()
...
+                });
+
+                Db.GetDocument("doc1").GetString("name").Should().Be("Jim");
+            }
+        }
+
+        [Fact]
+        public void TestConflictHandlerWithDeletedOldDoc()
+        {
+            using (var doc1 = new MutableDocument("doc1")) {
```
borrrden: This does not test "deleted old doc".  In order to do that you'd need to delete and then try to save with the conflict handler.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: I know it's confusing but curDoc and currDoc are different. I freed curDoc (a C4Document) on line 845 and 845.

djpongh: are the better names the curDoc and currDoc? I can easily see someone misreading one of the "r"'s and getting confused.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -816,10 +816,41 @@ public void RemoveChangeListener(ListenerToken token)
...
+        /// <summary>
+        /// Saves a document to the database. When write operations are executed concurrently, 
+        /// and if conflicts occur, conflict handler will be called. Use the handler to directly
+        /// edit the document.Returning true, will save the document. Returning false, will cancel
+        /// the save operation.
+        /// </summary>
+        /// <param name="document">The document to save</param>
+        /// <param name="conflictHandler">The conflict handler block which can be used to resolve it.</param> 
+        /// <returns><c>true</c> if the save succeeded, <c>false</c> if there was a conflict</returns>
+        public bool Save([NotNull]MutableDocument document, Func<MutableDocument, Document, bool> conflictHandler)
```
borrrden: `conflictHandler` should be `[NotNull]`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -816,10 +816,41 @@ public void RemoveChangeListener(ListenerToken token)
...
+        {
+            var doc = CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(document), document);
+            Debug.Assert(conflictHandler != null);
```
borrrden: Since this comes from outside the library, use `CBDebug.MustNotBeNull` the way that `doc` does.

PR #1153, opened 2019-05-13
Delete document revisions by doc_id, since we dont have sequence in docs table

kokoska69:
Docs table does not contain sequence, we cant select sequence and then delete corresponding revs.


borrrden:
I'm happy to have received a PR, but this is against a version that is past its end of life so what to do with it is unclear.  I will consult internally to decide on the appropriate course of action but it might end up with this PR being closed.

kokoska69:
This PR was reaction to user having the same problem as me. https://forums.couchbase.com/t/debugging-disapearing-rows-with-current-1-from-revs-table/15826/11 At this time is probably a better idea to use newer version 2.x

PR #1157, opened 2019-05-26
Update the purge doc by doc to throw not found error when trying to p…

Sandychuang8:
Update the purge doc by doc to throw not found error when trying to purge the doc no longer existed and add a test to verify.


borrrden:
The checks failed but it doesn't look like an issue caused by this PR.  Can you confirm that all the tests pass?

PR #1159, opened 2019-05-27
Change behavior of Purge by document instance

Sandychuang8:


Sandychuang8:
Not sure how to remove something that is committed and pushed. So removed the remote branch and update the PR again..

borrrden:
For future reference, you just make another commit reversing the change you made.

Issue #1150, opened 2019-05-07
Change to WebSocket timeout in LiteCore


pasin:
Per this couchbase/couchbase-lite-ios#2447, the WebScoket implementation in LiteCore has implemented the timeout. To avoid conflict or duplication, we should disable WebSocket (read/write) timeout from the platform code. If it cannot be disabled, we probably need to set it to be longer that what LiteCore is using.

borrrden:
Migrated to https://issues.couchbase.com/browse/CBL-39

PR #1160, opened 2019-05-30
#1150 Remove Websocket timeout

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -385,31 +375,24 @@ private void OnSocketReady()
...
-            cts.CancelAfter(IdleTimeout);
             if (NetworkStream == null) {
                 WriteLog.To.Sync.E(Tag, "Socket reported ready, but no network stream available!");
                 DidClose(C4WebSocketCloseCode.WebSocketCloseAbnormal, "Unexpected error in client logic");
                 return;
             }
 
-            NetworkStream.ReadTimeout = (int)IdleTimeout.TotalMilliseconds;
-            NetworkStream.WriteTimeout = (int)IdleTimeout.TotalMilliseconds;
-            NetworkStream.WriteAsync(httpData, 0, httpData.Length, cts.Token).ContinueWith(t =>
```
borrrden: This write still needs to happen, minus the cancellation.  This is the HTTP connection for the web socket handshake.

borrrden:
@snej We talked about this a little, but what do you think the behavior here should be?  Because right now if we go forward with the current behavior what will happen is:

1. 15 seconds pass and LiteCore "gives up"
2. CBL sees the retry count is 0, and so waits for 2 seconds and then starts again
3. 15 seconds pass and LiteCore "gives up"
4. CBL sees the retry count is 1, and so waits for 4 seconds and then starts again
5. 15 seconds pass and LiteCore "gives up"
6. CBL sees the retry count is 2, and if this is one-shot it now gives up

For continuous this process will never end.  However, if LiteCore were to return a non-transient error after a certain number of attempts then this cycle would be broken.

snej:
For a one-shot, then, the new behavior is then that CBL tries to connect for 51 seconds, during which it makes 3 attempts. That sounds pretty reasonable, and comparable to what a web browser does in a similar situation.

PR #1162, opened 2019-06-06
Feature/ccr api update

Sandychuang8:
Custom Conflict resolution update base in API updates.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -878,8 +878,7 @@ namespace Couchbase.Lite
             Debug.Assert(docID != null);
             var success = false;
             while (!success) {
-                Document localDoc = null, remoteDoc = null;
-                MutableDocument resolvedDoc = null;
+                Document localDoc = null, remoteDoc = null, resolvedDoc = null;
                 try {
                     localDoc = new Document(this, docID);
                     if (!localDoc.Exists) {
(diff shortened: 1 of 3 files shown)


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -899,29 +900,31 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
...
-                            else if (resolvedDoc.Database != this)
-                                throw new InvalidOperationException($"Resolved document db {resolvedDoc.Database.Name} is different from expected db {this.Name}");
-                        }
+                    var conflict = new Conflict(docID, localDoc.IsDeleted ? null : localDoc, remoteDoc.IsDeleted ? null : remoteDoc);
 
-                        InBatch(() =>
-                        {
-                            success = SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
-                        });
+                    resolvedDoc = conflictResolver.Resolve(conflict)?.ToMutable();
```
borrrden: This means that the original `resolvedDoc` will be undisposed.  It's not a bug, but better to try to dispose it.  Something like this will be better:

```c#
resolvedDoc = conflictResolver.Resolve(conflict);
Misc.SafeSwap(ref resolvedDoc, resolvedDoc?.ToMutable());
```

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -899,29 +900,31 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
...
+                    if (resolvedDoc != null) {
+                        if (resolvedDoc.Id != docID) {
+                            WriteLog.To.Sync.W(Tag, $"Resolved docID {resolvedDoc.Id} does not match docID {docID}",
+                                new SecureLogString(docID, LogMessageSensitivity.PotentiallyInsecure));
+                            resolvedDoc = new MutableDocument(docID, resolvedDoc.ToDictionary());
```
borrrden: You could use `SafeSwap` again here

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1276,7 +1276,7 @@ public void TestConflictResolverExceptionsWrongIDThrown()
                 return new MutableDocument("wrong_id");
             });
             
-            TestConflictResolverExceptionThrown(wrongDocIDResolver, false);
+            //TestConflictResolverExceptionThrown(wrongDocIDResolver, false);
```
borrrden: This test should be modified to ensure the correct behavior.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -899,29 +900,31 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
...
+                        } else if (resolvedDoc.Database == null)
+                            resolvedDoc.Database = this;
+                        else if (resolvedDoc.Database != this)
```
borrrden: Add brackets to all branches, even if only one line (style comment)

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
Sandychuang8: Yes, test name is wrong too. So this test will be removed. Additional tests will be added into another PR.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1269,30 +1269,21 @@ public void TestConflictResolverDeletedRemoteWin()
...
-        public void TestConflictResolverExceptionsWrongIDThrown()
-        {
-            var wrongDocIDResolver = new TestConflictResolver((conflict) =>
-            {
-                return new MutableDocument("wrong_id");
-            });
-            
-            TestConflictResolverExceptionThrown(wrongDocIDResolver, false);
-        }
-
```
borrrden: Don't remove this test, instead make sure it succeeds instead

borrrden:
Nevermind about my last review, I missed your comment about tests being added in another PR

PR #1164, opened 2019-06-08
Add additional conflict resolver tests

Sandychuang8:
Add additional conflict resolver tests: TestConflictResolverWrongDocID, TestConflictResolverReturningBlob, and TestConflictResolverReturningBlobFromDifferentDB

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-net/1164)

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -490,19 +490,13 @@ namespace Couchbase.Lite.Sync
                     try {
                         Config.Database.ResolveConflict(replication.Id, Config.ConflictResolver);
                         replication = replication.ClearError();
+                    } catch (CouchbaseException e) {
+                        replication.Error = e;
                     } catch (Exception e) {
-                        C4Error error;
-                        if (e.GetBaseException() is CouchbaseException) {
-                            var errCode = ((CouchbaseException)e).Error;
-                            error = new C4Error((C4ErrorCode)errCode);
-                            replication = replication.AssignError(error, (CouchbaseException)e);
-                        } else {
-                            error = new C4Error(C4ErrorCode.UnexpectedError);
-                            replication = replication.AssignError(error,
-                                new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e));
-                        }
-                        WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", e);
+                        replication.Error = new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e);
                     }
+                    if(replication.Error != null)
+                        WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", replication.Error);
                     _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, false));
                 });
                 _conflictTasks.TryAdd(t.ContinueWith(task => _conflictTasks.TryRemove(t, out var dummy)), 0);
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -284,7 +284,7 @@ namespace Couchbase.Lite.Sync
             }
 
             var replicatedDocumentsContainConflict = new List<ReplicatedDocument>();
-            var documentReplications = new List<ReplicatedDocument>((int)numDocs);
+            var documentReplications = new List<ReplicatedDocument>();
             for (int i = 0; i < (int) numDocs; i++) {
                 var current = docs[i];
                 if (!pushing && current->error.domain == C4ErrorDomain.LiteCoreDomain &&
@@ -298,15 +298,20 @@ namespace Couchbase.Lite.Sync
             }
 
             var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
-            replicator?.DispatchQueue.DispatchAsync(() =>
-            {
-                replicator.OnDocEnded(documentReplications, pushing);
-            });
 
-            replicator?.DispatchQueue.DispatchAsync(() =>
-            {
-                replicator.OnDocEndedWithConflict(replicatedDocumentsContainConflict, pushing);
-            });
+            if (documentReplications.Count > 0) {
+                replicator?.DispatchQueue.DispatchAsync(() =>
+                {
+                    replicator.OnDocEnded(documentReplications, pushing);
+                });
+            }
+
+            if (replicatedDocumentsContainConflict.Count > 0) {
+                replicator?.DispatchQueue.DispatchAsync(() =>
+                {
+                    replicator.OnDocEndedWithConflict(replicatedDocumentsContainConflict);
+                });
+            }
         }
 
         #if __IOS__
@@ -469,7 +474,7 @@ namespace Couchbase.Lite.Sync
             return true;
         }
 
-        private void OnDocEndedWithConflict(List<ReplicatedDocument> replications, bool pushing)
+        private void OnDocEndedWithConflict(List<ReplicatedDocument> replications)
         {
             if (_disposed) {
                 return;
@@ -494,11 +499,11 @@ namespace Couchbase.Lite.Sync
                         } else {
                             error = new C4Error(C4ErrorCode.UnexpectedError);
                             replication = replication.AssignError(error,
-                                new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message));
+                                new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e));
                         }
                         WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", e);
                     }
-                    _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, pushing));
+                    _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, false));
                 });
                 _conflictTasks.TryAdd(t.ContinueWith(task => _conflictTasks.TryRemove(t, out var dummy)), 0);
             }
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -79,7 +79,7 @@ namespace Couchbase.Lite.Sync
         private C4Replicator* _repl;
         private int _retryCount;
         private bool _stopping;
-        private ConcurrentBag<Task> _conflictTasks = new ConcurrentBag<Task>();
+        private ConcurrentDictionary<Task, int> _conflictTasks = new ConcurrentDictionary<Task, int>();
 
         #endregion
 
@@ -283,16 +283,14 @@ namespace Couchbase.Lite.Sync
                 return;
             }
 
-            var conflictTasks = new ConcurrentBag<Task>();
-            var replicatedDocuments = new ConcurrentBag<ReplicatedDocument>();
+            var replicatedDocumentsContainConflict = new List<ReplicatedDocument>();
             var documentReplications = new List<ReplicatedDocument>((int)numDocs);
             for (int i = 0; i < (int) numDocs; i++) {
                 var current = docs[i];
                 if (!pushing && current->error.domain == C4ErrorDomain.LiteCoreDomain &&
                     current->error.code == (int)C4ErrorCode.Conflict) {
-                    conflictTasks.Add(Task.Run(() =>
-                        replicatedDocuments.Add(new ReplicatedDocument(current->docID.CreateString() ?? "",
-                            current->flags, current->error, current->errorIsTransient))));
+                    replicatedDocumentsContainConflict.Add(new ReplicatedDocument(current->docID.CreateString() ?? "",
+                        current->flags, current->error, current->errorIsTransient));
                 } else {
                     documentReplications.Add(new ReplicatedDocument(current->docID.CreateString() ?? "",
                         current->flags, current->error, current->errorIsTransient));
@@ -305,10 +303,9 @@ namespace Couchbase.Lite.Sync
                 replicator.OnDocEnded(documentReplications, pushing);
             });
 
-            Task.WaitAll(conflictTasks.ToArray());
             replicator?.DispatchQueue.DispatchAsync(() =>
             {
-                replicator.OnDocEnded(replicatedDocuments, pushing);
+                replicator.OnDocEndedWithConflict(replicatedDocumentsContainConflict, pushing);
             });
         }
 
@@ -371,10 +368,9 @@ namespace Couchbase.Lite.Sync
             bool transient;
             if ((replicator != null && replicator.IsPermanentError(status.error, out transient)) 
                 && status.level == C4ReplicatorActivityLevel.Stopped) {
-                var array = replicator?._conflictTasks?.ToArray();
+                var array = replicator?._conflictTasks?.Keys?.ToArray();
                 if (array != null) {
                     Task.WaitAll(array);
-                    replicator._conflictTasks = new ConcurrentBag<Task>();
                 }
             }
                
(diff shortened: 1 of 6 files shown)


src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1161,7 +1164,10 @@ public void TestConflictResolverMergeDoc()
                 _otherDB.Save(doc1aMutable);
             }
 
-            RunReplication(config, 0, 0);
+            RunReplication(config, 0, 0, documentReplicated: (sender, args) =>
+            {
+                conflictResolved.Should().Be(true, "Because the DocumentReplicationEvent will not be notified until the conflicted document is resolved.");
+            });
```
borrrden: This is not going to be needed.  All resolutions are required to finish before replicator stop.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1291,6 +1319,107 @@ public void TestConflictResolverExceptionsNoneMatchDBThrown()
...
+            Db.Count.ShouldBeEquivalentTo(1);
+
+            using (var doc1a = _otherDB.GetDocument("doc1"))
+            using (var doc1aMutable = doc1a.ToMutable()) {
+                var luckyByteArray = new byte[] { 7, 7, 7 };
+                doc1aMutable.SetBlob("blob", new Blob("text/plaintext", luckyByteArray));
+                _otherDB.Save(doc1aMutable);
+            }
+        }
+
```
borrrden: This seems like a lot of duplication.  Is there a way to refactor the original method to optionally use blobs?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1291,6 +1319,107 @@ public void TestConflictResolverExceptionsNoneMatchDBThrown()
...
+        [Fact]
+        public void TestConflictResolverReturningBlobFromDifferentDB()
+        {
+            var tmpDoc = new MutableDocument("doc1");
+            var evilByteArray = new byte[] { 6, 6, 6 };
+            using (var thirdDb = new Database("different_db")) {
+                tmpDoc.SetBlob("blob", new Blob("text/plaintext", evilByteArray));
+                thirdDb.Save(tmpDoc);
+
+                var differentDbResolver = new TestConflictResolver((conflict) => tmpDoc);
```
borrrden: This isn't quite correct.  You need to make a new document with the blob from another DB.  If it's saved in a different DB then this is the same test as the differing DB test.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1291,6 +1313,51 @@ public void TestConflictResolverExceptionsNoneMatchDBThrown()
...
+        [Fact]
+        public void TestConflictResolverReturningBlobFromDifferentDB()
+        {
+            var evilByteArray = new byte[] { 6, 6, 6 };
+            using (var thirdDb = new Database("different_db"))
+            using (var doc1 = new MutableDocument("doc1")) {
+                doc1.SetBlob("blob", new Blob("text/plaintext", evilByteArray));
+                thirdDb.Save(doc1);
+
+                var differentDbResolver = new TestConflictResolver((conflict) => doc1);
```
borrrden: Still this is returning a document from another database.  You need to create a new document *inside the conflict resolver* and set a blob on it that comes from a document from another database.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -482,9 +481,11 @@ private void OnDocEnded(List<ReplicatedDocument> replications, bool pushing)
                             Config.Database.ResolveConflict(docID, Config.ConflictResolver);
                             replication = replication.ClearError();
                         } catch (Exception e) {
+                            replication = replication.AssignError(new C4Error(C4ErrorCode.NotFound));
```
borrrden: This is going to assign "NotFound" to any exception thrown unconditionally.  See what happens now if you try to verify the errors on other kinds of errors like "db from different document", etc

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -472,7 +472,6 @@ private void OnDocEnded(List<ReplicatedDocument> replications, bool pushing)
                 var logDocID = new SecureLogString(docID, LogMessageSensitivity.PotentiallyInsecure);
                 if (!pushing && error.domain == C4ErrorDomain.LiteCoreDomain &&
                     error.code == (int)C4ErrorCode.Conflict) {
-                    replications.Remove(replication);
```
borrrden: This could easily result in two document ended events for a conflict.  Once with "conflict" and one with no error after the conflict resolution is done.  This needs to stay here.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -283,11 +283,20 @@ private static void OnDocEnded(C4Replicator* repl, bool pushing, IntPtr numDocs,
...
             var documentReplications = new List<ReplicatedDocument>((int)numDocs);
             for (int i = 0; i < (int) numDocs; i++) {
                 var current = docs[i];
-                documentReplications.Add(new ReplicatedDocument(current->docID.CreateString() ?? "", 
-                    current->flags, current->error, current->errorIsTransient));
+                if (!pushing && current->error.domain == C4ErrorDomain.LiteCoreDomain &&
+                    current->error.code == (int)C4ErrorCode.Conflict) {
+                    conflictTasks.Add(Task.Run(() =>
+                        replicatedDocuments.Add(new ReplicatedDocument(current->docID.CreateString() ?? "",
+                            current->flags, current->error, current->errorIsTransient))));
```
borrrden: There is no reason for the above to be asynchronous.  It should just be a simple add to a collection.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +473,53 @@ private bool HandleError(C4Error error)
...
+                _conflictTasks.Add(Task.Run(() =>
+                {
+                    ReplicatedDocument replication;
+                    if (replications.TryTake(out replication)) {
+                        try {
+                            Config.Database.ResolveConflict(replication.Id, Config.ConflictResolver);
+                            replication = replication.ClearError();
+                        } catch (Exception e) {
+                            C4Error error;
+                            if(e.GetBaseException() is CouchbaseException) {
```
borrrden: Only CouchbaseLiteException will have C4 error codes.  The rest of the Couchbase exception family has different domain codes.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +473,53 @@ private bool HandleError(C4Error error)
...
+                                var errCode = ((CouchbaseException)e).Error;
+                                error = new C4Error((C4ErrorCode)errCode);
+                            } else {
+                                error = new C4Error(C4ErrorCode.UnexpectedError);
+                            }
+                            replication = replication.AssignError(error);
```
borrrden: It would be good to preserve the original exception's message here.  The tests should be checking this.  Consider setting the error *and* the exception by altering the `ReplicatedDocument` class

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -357,9 +371,10 @@ private static void StatusChangedCallback(C4Replicator* repl, C4ReplicatorStatus
             bool transient;
             if ((replicator != null && replicator.IsPermanentError(status.error, out transient)) 
                 && status.level == C4ReplicatorActivityLevel.Stopped) {
-                var array = replicator?._conflictTasks?.Keys?.ToArray();
+                var array = replicator?._conflictTasks?.ToArray();
                 if (array != null) {
                     Task.WaitAll(array);
+                    replicator._conflictTasks = new ConcurrentBag<Task>();
```
borrrden: This new implementation means that this bag will never be cleared unless a replicator stops.  So continuous replication will grow unbounded.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +473,53 @@ private bool HandleError(C4Error error)
...
 
-        private void OnDocEnded(List<ReplicatedDocument> replications, bool pushing)
+        private void OnDocEnded(ConcurrentBag<ReplicatedDocument> replications, bool pushing)
         {
             if (_disposed) {
                 return;
             }
 
-            for(int i = 0; i < replications.Count; i++) {
+            while (!replications.IsEmpty) {
```
borrrden: This is not a good loop.  It will keep looping adding in tasks that will eventually do nothing at the end because its termination condition is delayed.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +473,53 @@ private bool HandleError(C4Error error)
...
+                                var errCode = ((CouchbaseException)e).Error;
+                                error = new C4Error((C4ErrorCode)errCode);
+                            } else {
+                                error = new C4Error(C4ErrorCode.UnexpectedError);
+                            }
+                            replication = replication.AssignError(error);
+                            WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", e);
+                        }
+                    }
+                    _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, pushing));
```
borrrden: This will fire regardless of whether or not the above `TryTake` succeeded or failed.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +473,53 @@ private bool HandleError(C4Error error)
...
+                }).ContinueWith(task => _conflictTasks.TryTake(out var t)));
```
borrrden: `TryTake` does not do what you think it does.  This will remove a random element from the bag, so probably you've just skipped things by removing a different task.  This is the reason for making it a concurrent dictionary (there is no way to remove a specific item from a concurrent bag)

Sandychuang8: Are we able to reset _conflictTasks after waitAll call?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: this is outdated.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
```diff
@@ -70,28 +71,37 @@ public struct ReplicatedDocument
         [CanBeNull]
         public CouchbaseException Error { get; }
 
+        internal Exception Exception { get; set; }
```
borrrden: How does this help the user see the original message?  There needs to be a way to preserve the original exception in a way that is visible to the user (e.g. `Error`'s `InnerException`)

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -296,6 +303,10 @@ private static void OnDocEnded(C4Replicator* repl, bool pushing, IntPtr numDocs,
                 replicator.OnDocEnded(documentReplications, pushing);
             });
 
+            replicator?.DispatchQueue.DispatchAsync(() =>
+            {
+                replicator.OnDocEndedWithConflict(replicatedDocumentsContainConflict, pushing);
+            });
```
borrrden: It might be better to have both of these calls inside of one dispatch async instead of two, but it's up to you.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -458,42 +469,57 @@ private bool HandleError(C4Error error)
             return true;
         }
 
-        private void OnDocEnded(List<ReplicatedDocument> replications, bool pushing)
+        private void OnDocEndedWithConflict(List<ReplicatedDocument> replications, bool pushing)
```
borrrden: FYI:  `pushing` will always be `true` here

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
```diff
@@ -71,14 +71,14 @@ public struct ReplicatedDocument
         [CanBeNull]
         public CouchbaseException Error { get; }
 
-        internal Exception Exception { get; set; }
+        internal CouchbaseLiteException Exception { get; set; }
```
borrrden: Still this is not helpful because it's **internal**.  The public facing `Error` is where you want to record this information so that the user can actually see it.  The `Error` should be as follows:

1. Original exception is CouchbaseException (or derived) -> Set `Error` directly
2. Original exception is not CouchbaseException (or derived) -> Create a new `CouchbaseLiteException` with the original exception as `InnerException` and set `Error` to that
3. No original exception, just error code -> Use `CouchbaseException.Create()`

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
```diff
@@ -71,7 +71,7 @@ public struct ReplicatedDocument
         [CanBeNull]
         public CouchbaseException Error { get; }
 
-        internal CouchbaseLiteException Exception { get; set; }
+        public CouchbaseLiteException Exception { get; }
```
borrrden: I think I need to be clearer here.  Get rid of this property.  Everything should be on the `Error` property only.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -492,10 +492,10 @@ private void OnDocEndedWithConflict(List<ReplicatedDocument> replications)
                         replication = replication.ClearError();
                     } catch (Exception e) {
                         C4Error error;
-                        if (e.GetBaseException() is CouchbaseLiteException) {
-                            var errCode = ((CouchbaseLiteException)e).Error;
+                        if (e.GetBaseException() is CouchbaseException) {
+                            var errCode = ((CouchbaseException)e).Error;
                             error = new C4Error((C4ErrorCode)errCode);
```
borrrden: This has taken a step backwards.  This will end up reporting incorrect error codes.  I didn't want to outright suggest code blocks but perhaps you should try this:

Make `ReplicatedDocument.Error` settable interally

```
} catch (CouchbaseException e) {
    replication.Error = e;
} catch (Exception e) {
    replication.Error = new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e);
}
```

Then you can put the `ReplicatedDocument` constructors back the way they were.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -490,19 +490,13 @@ private void OnDocEndedWithConflict(List<ReplicatedDocument> replications)
...
-                        } else {
-                            error = new C4Error(C4ErrorCode.UnexpectedError);
-                            replication = replication.AssignError(error,
-                                new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e));
-                        }
-                        WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", e);
+                        replication.Error = new CouchbaseLiteException(C4ErrorCode.UnexpectedError, e.Message, e);
                     }
+                    if(replication.Error != null)
+                        WriteLog.To.Sync.W(Tag, $"Conflict resolution of '{replication.Id}' failed", replication.Error);
```
borrrden: Confirm with Pasin but I think this logging statement is not needed.

Sandychuang8: @pasin said the chance logging this warning is very small but it's good to log it in case we encounter any edge case..
So far I see iOS has it `https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLReplicator.mm#L599` @bmeike probably also has it... If not Blake, please log a warning message for exception if any.

bmeike: I have that same log message in my code.  It's been there forever...

bmeike: I log resolver failure, wrong db and wrong id.  The first two also get returned to the client code, as exceptions (they are returned, not thrown)

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1275,7 +1280,225 @@ public void TestConflictResolverDeletedRemoteWin()
...
+                doc.SetString("wrong_id_key", "wrong_id_value");
+                return doc;
+            });
+
+            RunReplication(config, 0, 0);
+
+            using (var db = Db.GetDocument("doc1")) {
+                db.GetString("wrong_id_key").Should().Be("wrong_id_value");
+            }
+        }
```
bmeike: Do you want to verify that this doesn't return an error to the user?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1275,7 +1280,225 @@ public void TestConflictResolverDeletedRemoteWin()
...
+            config.ConflictResolver = new TestConflictResolver((conflict) =>
+            {
+                var cnt = 0;
+                lock (q) {
+                    q.Enqueue(conflict.LocalDocument.Id);
+                    cnt = q.Count;
+                }
+
+                if (cnt == 1) {
+                    Thread.Sleep(500);
```
bmeike: This is kinda flakey.  Does iOS have latches of some kind, so that you can wait for thee other guy to finish, instead of waiting for time?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1275,7 +1280,225 @@ public void TestConflictResolverDeletedRemoteWin()
...
+                    }
+
+                    return args.Status.Activity == ReplicatorActivityLevel.Stopped;
+                });
+            });
+
+            replicator.Start();
+           
+            Thread.Sleep(50);
+            replicator1.Start();
```
bmeike: Again... hate those Sleeps.  Isn't there a way to do that deterministically?

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
borrrden: Yes, this is not an error condition since we correct it in the code base.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
borrrden: Not this one, it's a straight up hack around the BLIP issue that I mentioned yesterday.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
borrrden: Yeah, that's a valid improvement.  Here is how it could play out:

`ManualResetEvent` outside of the callback.  If `cnt` is 1 then block on it, otherwise ignore it.  Do the rest of the logic, then if `cnt` is *not* 1 signal the event.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1374,8 +1375,11 @@ public void TestNonBlockingConflictResolver()
                 }
 
                 if (cnt == 1) {
-                    Thread.Sleep(500);
+                    manualResetEvent.WaitOne();
+                } else {
+                    manualResetEvent.Set();
```
borrrden: Call `Set()` later, before the return because the guarantee is supposed to be "both of the second resolver's enqueue calls are done"

PR #1173, opened 2019-08-08
Mutable Fleece array and dictionary bindings

Sandychuang8:


borrrden:
I'd like this to go hand in hand with code that actually uses it.

PR #1174, opened 2019-08-14
CBL-247 Add Jenkinsfile to allow PR validation

borrrden:


borrrden:
Need to close this until later because it is still based on packages

PR #1178, opened 2019-08-28
CBL-328 Use new c4repl_start API

Sandychuang8:


borrrden:
Don't forget to put the JIRA numbers in your commits and PRs so that JIRA will track their progress.

Sandychuang8:
The current unit tests are all passing. I wonder if I should add additional unit tests to verify the new API?

borrrden:
If the current tests are passing that is good enough since this PR is so small and directly wrapped around LiteCore (which tests this API already)

PR #1179, opened 2019-09-03
Error messages constants

Sandychuang8:


Sandychuang8:
@bmeike @jayahariv @pasin 
Please review src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs 
after it's approved, please make a similar addition to your platform source and from now on we will be add all the public facing error messages here so we can standardize the error messages..

pasin:
@jayahariv please review this for iOS.

Sandychuang8:
> Shouldn't these strings go into some kind of lazily acquired resource bundle?

Isn't that's more for app development to support localization?

bmeike:
That is one reason to do it.  Another is to separate data from code.  If this were in some reasonably portable format, we could all (platforms) use *exactly* the same data.  That seem like a pretty worthwhile goal...

bmeike:
Seriously, folks... Can we put this stuff in a JSON/XML file somewhere?  If you don't want the bindings to read it directly from the common file, can we at least have some common source from which it is easy to auto-generate each platform's version of these messages?

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+        //MArray MDict <--- in the future, we will replace them with LiteCore Mutable Fleece API
+        internal const string CannotRemoveItemsFromNonMutableMArray = "Cannot remove items from a non-mutable MArray";
+        internal const string CannotRemoveStartingFromIndexLessThan = "Cannot remove starting from an index less than 0 (got {0})";
+        internal const string CannotRemoveRangeEndsBeforeItStarts = "Cannot remove a range that ends before it starts (got start= {0}, count = {1} )";
+        internal const string RangeEndForRemoveExceedsArrayLength = "Range end for remove exceeds the length of the array(got start = {0}, count = {1} )";
+        internal const string CannotSetItemsInNonMutableMArray = "Cannot set items in a non-mutable MArray";
+        internal const string CannotClearNonMutableMArray = "Cannot clear a non-mutable MArray";
+        internal const string CannotInsertItemsInNonMutableMArray = "Cannot insert items in a non-mutable MArray";
+        internal const string CannotClearNonMutableMDict = "Cannot clear a non-mutable MDict";
+        internal const string CannotSetItemsInNonMutableInMDict = "Cannot set items in a non-mutable MDict";
```
Sandychuang8: @bmeike The bundle idea is good, but I feel it is overkill for just less than 50 error messages. Also in the future the number of error messages will go further down because some of them will be coming from LiteCore instead.... But let's vote for it and check the doability of it. cc. @jayahariv

bmeike: Ok.  What does "overkill" mean?  Is it too hard to maintain?  Is there some downside that I've missed?

Sandychuang8: The xml schema for each platform will not be exactly the same, so some modification will be needed. So out of box shareable is not possible.

Sandychuang8: For .Net, I believe I need to prepare different ways to support Resource Bundle for each platforms as well.

bmeike: Why?  I cannot see any reason that that would be the case...

Sandychuang8: The xml schema for each platform will not be exactly the same.

bmeike: I hear you say that, but I don't understand why it is true.  Why could they not all be the same?  At the very worst, we could write little perl/Python programs that generated per-platform implementations from the common source of truth...

Sandychuang8: The resource bundle xml (xaml for .Net) schema are not be the same. I know from my past experience.....

bmeike: Got it.  That is easy to fix:
- Don't use a standard resource bundle.  Some small simple parsing (you already have a JSON parser...) would do it.
- Like I suggested, autogenerate the per-platform resource bundles from a single source of truth.  It would take, like 20 lines of perl or python

borrrden: If we are autogenerating then why not just autogenerate the file above from our data source.  Adding in a resource bundles when we have the ability to generate per platform source simply adds complexity.  I bet I could even write a little .NET program that transforms the C# file into JSON or XML or YAML or even INI.

bmeike: RIght.  I don't care if the source of truth is one of the platform native files.  This one is annoying but not impossible to parse.  Though I can't see any reason we'd convert it to XML or YAML.  Nobody can use those.  I just want a single source of error messages.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Text;
+
+namespace Couchbase.Lite
+{
+    internal static class CouchbaseLiteErrorMessage
+    {
+        //Database error messages
+        internal const string CreateDBDirectoryFailed = "Unable to create database directory";
+        //Database - Copy
+        internal const string ResolveDefaultDirectoryFailed = "Failed to resolve a default directory! If you have overriden the IDefaultDirectoryResolver interface, please check it.  Otherwise please file a bug report.";
```
jayahariv:
```suggestion
        internal const string ResolveDefaultDirectoryFailed = "Failed to resolve a default directory! If you have overriden the default directory, please check it.  Otherwise please file a bug report.";
```

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+        internal const string InvalidPath = "Path.Combine failed to return a non-null value!";
```
jayahariv: Path.Combine?

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+        //Database - Purge
+        internal const string DocumentNotFound = "Document doesn't exist in the database.";
+        internal const string DocumentAnotherDatabase = "Cannot operate on a document from another database";
+        //Database - Save
+        internal const string BlobDifferentDatabase = "A document contains a blob that was saved to a different database; the save operation cannot complete.";
+        internal const string BlobContentNull = "No data available to write for install. Please ensure that all blobs in the document have non-null content.";
+        //Database - Resolve Conflict
+        internal const string ResolvedDocContainsNull = "Resolved document contains a null body";
+        internal const string ResolvedDocFailedLiteCore = "LiteCore failed resolving conflict.";
+        internal const string ResolvedDocWrongDb = "Resolved document db {0} is different from expected db {1}";
```
jayahariv: {0}  = is this a placeholder or just string, pointing to the first argument printed after this line.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
borrrden: This is a .NET specific exception that I put in to silence a ReSharper warning.  It doesn't need to be in this list.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
borrrden: It's a format placeholder.  The first and second arguments passed to `String.Format` on .NET will be used in the final message.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+        internal const string DBClosed = "Attempt to perform an operation on a closed database";
+        internal const string NoDocumentRevision = "No revision data on the document!";
+        //Fragment - Value
+        internal const string FragmentPathNotExist = "Specified fragment path does not exist in object, cannot set value";
+        //ToCouchbaseObject(object value))
+        internal const string InvalidCouchbaseObjType = "{0} is not a valid type. " +
+                                                "You may only pass byte, sbyte, short, ushort, int, uint, long, ulong, float, double, bool, DateTimeOffset, Blob, " +
```
jayahariv: in case of iOS, `Data, Int, Int64, Float, Double, String, Date, Bool, Blob, Any`

how about `You may only pass language supported types`?

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
jayahariv: is the compiler generate respective cross platform code placeholders for us.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -0,0 +1,84 @@
...
+        //Query QueryResultSet
+        internal const string ResultSetAlreadyEnumerated = "This result set has already been enumerated, please re-run Execute() on the original query";
+        //Query
+        internal const string ExpressionsMustContainOnePlusElement = "{0} expressions must contain at least one element";
+        internal const string DuplicateSelectResultName = "Duplicate select result named {0}";
+        internal const string NoAliasInJoin = @"The default database must have an alias in order to use a JOIN statement 
+                                                (Make sure your data source uses the As() function)";
+        internal const string InvalidQueryDBNull = "Invalid query, Database == null";
+        internal const string InvalidQueryMissingSelectOrFrom = "Invalid query, missing Select or From";
+        //MArray MDict <--- in the future, we will replace them with LiteCore Mutable Fleece API
```
jayahariv: assume below are MArray, MDict, which iOS doesn't use.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
borrrden: Too vague.  Maybe this needs to be another formatted message.

src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
borrrden: Yes, I think it should.

borrrden:
This PR should not be merged since the actual file that we will use is going to come from a tool developed elsewhere.  However, we should not include stuff that isn't going to be used by *every* platform, so let's take out the platform mutable fleece stuff.

borrrden:
You should have received the generated list now, so that should be a part of this PR instead.

PR #1180, opened 2019-09-09
CBL-349 Catching exception thrown when conflictResolver.Resolve failed

Sandychuang8:
Fixes CBL-349 by catching exception when running conflictResolver.Resolve failed. Also fix unit test TestNonBlockingDatabaseOperationConflictResolver.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -934,9 +934,6 @@ namespace Couchbase.Lite
                         // ReSharper disable once AccessToDisposedClosure
                         writeSuccess = SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
                     });
-                    
-                } catch (Exception ex) {
-                    throw ex;
                 } finally {
                     resolvedDoc?.Dispose();
                     localDoc?.Dispose();
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -935,6 +935,8 @@ namespace Couchbase.Lite
                         writeSuccess = SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
                     });
                     
+                } catch (Exception ex) {
+                    throw ex;
                 } finally {
                     resolvedDoc?.Dispose();
                     localDoc?.Dispose();
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -1329,23 +1329,6 @@ namespace Test
         }
 
         [Fact]
-        public void TestExceptionThrownInConflictResolver()
-        {
-            CreateReplicationConflict("doc1");
-            var config = CreateConfig(false, true, false);
-            config.ConflictResolver = new TestConflictResolver((conflict) =>
-            {
-                    using (var d = Db.GetDocument("doc1"))
-                    using (var doc = d.ToMutable()) {
-                        d.GetString("name").Should().Be("Tiger");
-                    }
-                return null;
-            });
-
-            RunReplication(config, 0, 0, isConflictResolvingFailed: true);
-        }
-
-        [Fact]
         public void TestNonBlockingDatabaseOperationConflictResolver()
         {
             int resolveCnt = 0;
@@ -1356,7 +1339,7 @@ namespace Test
                 if (resolveCnt == 0) {
                     using (var d = Db.GetDocument("doc1"))
                     using (var doc = d.ToMutable()) {
-                        d.GetString("name").Should().Be("Cat");
+                        d.GetString("name").Should().Be("WrongAnswer");
                         doc.SetString("name", "Cougar");
                         Db.Save(doc);
                         doc.GetString("name").Should().Be("Cougar", "Because database save operation was not blocked");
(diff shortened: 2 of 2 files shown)


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -935,6 +935,8 @@ internal void ResolveConflict([NotNull]string docID, [CanBeNull]IConflictResolve
                         writeSuccess = SaveResolvedDocument(resolvedDoc, localDoc, remoteDoc);
                     });
                     
+                } catch (Exception ex) {
+                    throw ex;
```
borrrden: This doesn't change the behavior.  If there is no `catch` then the exception is thrown by default.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1339,10 +1339,10 @@ public void TestNonBlockingDatabaseOperationConflictResolver()
                 if (resolveCnt == 0) {
                     using (var d = Db.GetDocument("doc1"))
                     using (var doc = d.ToMutable()) {
-                        d.GetString("name").Should().Be("Tiger");
+                        d.GetString("name").Should().Be("Cat");
                         doc.SetString("name", "Cougar");
                         Db.Save(doc);
-                        d.GetString("name").Should().Be("Cougar", "Because database save operation was not blocked");
+                        doc.GetString("name").Should().Be("Cougar", "Because database save operation was not blocked");
```
borrrden: It's good to fix the exception, but this is not the only thing that needs to happen.  What happens if this fails?  How will the test know?  If an exception is thrown inside a conflict resolver then the exception is swallowed, logged, and the resolution is aborted. This test still passes in this case, as it was passing before even though this logic was wrong.  That needs to be corrected.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1930,6 +1947,20 @@ private void VerifyChange(ReplicatorStatusChangedEventArgs change, int errorCode
...
+                        foreach(var d in args.Documents) {
+                            d.Error.Should().NotBeNull();
+                            var error = d.Error.As<CouchbaseException>();
+                            error.Error.Should().BeGreaterThan(0);
+                            error.Domain.Should().NotBeNull();
+                        }
+                    }
+                });
+            }
+
```
borrrden: This is not what I had in mind....once again this is throwing exceptions (via asserts) inside of a callback.  This passes now, but there is no way to detect if it fails....

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1328,6 +1328,23 @@ public void TestConflictResolverCalledTwice()
...
+            {
+                    using (var d = Db.GetDocument("doc1"))
+                    using (var doc = d.ToMutable()) {
+                        d.GetString("name").Should().Be("Tiger");
+                    }
+                return null;
+            });
+
+            RunReplication(config, 0, 0, isConflictResolvingFailed: true);
+        }
```
borrrden: There is no need for a new test

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1339,24 +1338,36 @@ public void TestNonBlockingDatabaseOperationConflictResolver()
                 if (resolveCnt == 0) {
                     using (var d = Db.GetDocument("doc1"))
                     using (var doc = d.ToMutable()) {
-                        d.GetString("name").Should().Be("Tiger");
+                        d.GetString("name").Should().Be("Cat");
                         doc.SetString("name", "Cougar");
                         Db.Save(doc);
-                        d.GetString("name").Should().Be("Cougar", "Because database save operation was not blocked");
+                        doc.GetString("name").Should().Be("Cougar", "Because database save operation was not blocked");
```
borrrden: Not a problem with this PR in particular, but while you are fixing this test, this check doesn't actually check a proper save, it just checks that the in memory value was updated.  After the save, retrieve the document again and check the string to get a proper validation.

src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1339,24 +1338,36 @@ public void TestNonBlockingDatabaseOperationConflictResolver()
...
+                    foreach (var d in args.Documents) {
+                        _waitAssert1.RunConditionalAssert(() =>
+                        {
+                            return d.Error == null;
+                        });
+                    }
+                }
+            });
 
+            _waitAssert1.WaitForResult(TimeSpan.FromSeconds(2));
```
borrrden: This will work, but it won't tell you much about the error.  Remember that these need to be maintained and explained to the next person who works on them so simpler is better.  Perhaps something like:

```
Exception innerException = null;
config.ConflictResolver = new TestConflictResolver((conflict) =>
{
    try {
        // ...
    } catch(Exception e) {
        innerException = e;
    }
});

innerException.Should.BeNull();
```

Or in this case, there is something even simpler you can do.  You can ignore the above, let the exception throw, and notice that `resolveCnt` does not change in that case because it's below the part that throws the exception.  So by that logic, the last part of this should be:

```
// This will be 0 if the test resolver threw an exception
resolveCnt.Should().NotBe(0, "because otherwise the conflict resolver didn't complete"); 
using (var doc = Db.GetDocument("doc1")) {
    // Remove the if here, it doesn't add anything (and it's incorrect)
    doc.Should().BeNull();
}
```

PR #1182, opened 2019-09-30
CBL-267 Mutable fleece array refactor

Sandychuang8:
CBL-267 Mutable fleece array refactor


src/Couchbase.Lite.Shared/API/Document/ArrayObject.cs
```diff
@@ -20,7 +20,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-
+using Couchbase.Lite.Fleece;
```
borrrden: Preserve the spacing between .NET namespaces and others

src/Couchbase.Lite.Shared/API/Document/MutableArrayObject.cs
```diff
@@ -22,7 +22,7 @@
 using System;
 using System.Collections;
 using System.Linq;
-
+using Couchbase.Lite.Fleece;
```
borrrden: Same here for spacing

src/Couchbase.Lite.Shared/Serialization/FLValueConverter.cs
```diff
@@ -21,7 +21,7 @@
 using System.Collections.Generic;
 using System.Reflection;
 using System.Text;
-
+using Couchbase.Lite.Fleece;
```
borrrden: Also here

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -15,20 +15,29 @@
...
 using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
+using System.Text;
 
-using JetBrains.Annotations;
-
-using LiteCore.Interop;
-
-namespace Couchbase.Lite.Internal.Serialization
+namespace Couchbase.Lite.Fleece
```
borrrden: Group together similar usings (.NET framework, then others)

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -39,12 +48,16 @@ internal sealed unsafe class MArray : MCollection, IList<object>
 
         #region Properties
 
-        public FLArray* BaseArray { get; private set; }
+        private FLMutableArray* ar { get; set; }
```
borrrden: Name this something more meaningful

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -39,12 +48,16 @@ internal sealed unsafe class MArray : MCollection, IList<object>
...
+
+        private FLArray* BaseArray => (FLArray*)ar;
```
borrrden: Upon looking at this, I think it might be better to have one `_flArray` variable of type `FLArray *` and simply cast as needed.  I thought this would be two classes, but it is just one (which is better).

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -135,26 +149,37 @@ public void Set(int index, object val)
...
+            if (index < Count) {
+                PopulateArr();
+            }
+
             Mutate();
+            using (var encoded = val.FLEncode()) { 
+                //Convert object into FLValue
+                var flValue = NativeRaw.FLValue_FromData((FLSlice)encoded, FLTrust.Trusted);
+                Native.FLSlot_SetValue(Native.FLMutableArray_Set(ar, (uint)index), flValue);
+            }
```
borrrden: This will probably need to be reworked later.  There is a whole suite of `FLSlot_Set*` functions that are available and can be used by writing a method similar to `FLEncode()` that calls the appropriate `FLSlot_Set*` methods.  This avoids encoding upfront.

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -227,6 +255,44 @@ public bool Remove(object item)
...
         #endregion
 
+        #region IList<object>
+
+        public int IndexOf(object item)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Insert(int index, object val)
```
borrrden: This whole method looks like it is copied from `Set`, so it would be good to find a way to call it.  It looks like it is simply just `Set` with the addition of `Native.FLMutableArray_Insert(ar, (uint)index, 1);`

src/LiteCore/src/LiteCore.Shared/Interop/Fleece.cs
```diff
@@ -391,14 +391,26 @@ public static void FLEncode(this object obj, FLEncoder* enc)
...
                 case uint u:
-                    ((ulong) u).FLEncode(enc);
+                    ((ulong)u).FLEncode(enc);
                     break;
                 case ulong u:
                     u.FLEncode(enc);
                     break;
                 case int i:
-                    ((long) i).FLEncode(enc);
+                    ((long)i).FLEncode(enc);
```
borrrden: You can simplify these cases in two groups by doing the following:

```
case byte b:
case ushort us:
case uint ui:
case ulong ul:
    var numericVal = Convert.ToUInt64(obj);
    numericVal.FLEncode(enc);
```

And the signed version

src/LiteCore/src/LiteCore.Shared/Interop/Fleece_native.cs
```diff
@@ -198,6 +277,50 @@ public static string FLDictKey_GetString(FLDictKey* dictKey)
...
+        public static extern FLMutableDict* FLMutableDict_New();
+
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        public static extern FLDict* FLMutableDict_GetSource(FLMutableDict* x);
+
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        [return: MarshalAs(UnmanagedType.U1)]
+        public static extern bool FLMutableDict_IsChanged(FLMutableDict* x);
+
+        public static FLSlot* FLMutableDict_Set(FLMutableDict* FL_NONNULL, string key)
```
borrrden: This is really funny...I'll need to fix this script later to ignore the `FL_NONNULL` macro

src/LiteCore/src/LiteCore.Shared/Interop/Fleece_native.cs
```diff
@@ -134,9 +134,51 @@ public static string FLValue_ToString(FLValue* value)
...
+
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        public static extern void FLSlot_SetValue(FLSlot* x, FLValue* value);
+
         [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
         public static extern uint FLArray_Count(FLArray* array);
 
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        [return: MarshalAs(UnmanagedType.U1)]
+        public static extern bool FLArray_IsEmpty(FLArray* array);
```
borrrden: If you are going to bring this one in, bring in the dictionary version as well.

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -39,12 +45,14 @@ internal sealed unsafe class MArray : MCollection, IList<object>
 
         #region Properties
 
-        public FLArray* BaseArray { get; private set; }
+        private FLMutableArray* flArr { get; set; }
```
borrrden: No need to be a property, an instance variable is fine (`_flArr`).

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -167,30 +178,52 @@ private void Resize(int newSize)
...
+        {
+            if (index < Count) {
+                PopulateArr();
+            }
+
+            Mutate();
+            using (var encoded = val.FLEncode()) {
+                //Convert object into FLValue
+                var flValue = NativeRaw.FLValue_FromData((FLSlice)encoded, FLTrust.Trusted);
+                if(isInserting)
```
borrrden: Braces

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -227,6 +261,34 @@ public bool Remove(object item)
...
+        #region IList<object>
+
+        public int IndexOf(object item)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Insert(int index, object val)
+        {
+            if (!IsMutable) {
```
borrrden: I wonder if `IsMutable` will ever be `false` anymore since it cannot store a non mutable array....Can you find a situation like that?

borrrden: I was asking in general, not just about this line.  If it is not used then the property itself can be removed but I have a feeling that somehow this property needs to protect against writing into an array that is supposed to be immutable.

Sandychuang8: Looks like `IsMutable` is never be false... I don't see any public way to change the IsMutable to false.. However, the property is used in `public bool IsReadOnly => !IsMutable;` and IsReadOnly cannot be removed because it's the member of IList.

borrrden: It can get changed via `InitAsCopyOf` or `InitInSlot` from what I can tell so let's leave it in to be safe.

PR #1183, opened 2019-10-03
CBL-374 Mutable Fleece dictionary refactor

Sandychuang8:


src/Couchbase.Lite.Shared/API/Document/ArrayObject.cs
```diff
@@ -223,9 +224,15 @@ IEnumerator IEnumerable.GetEnumerator()
 
         #region IDisposable
 
+        /// <summary>
+        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+        /// </summary>
```
borrrden: Use `/// <inheritdoc />`

src/Couchbase.Lite.Shared/API/Document/DictionaryObject.cs
```diff
@@ -282,6 +284,21 @@ IEnumerator IEnumerable.GetEnumerator()
 
         #endregion
 
+        #region IDisposable
+
+        /// <summary>
+        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+        /// </summary>
```
borrrden: Use `/// <inheritdoc />`

src/Couchbase.Lite.Shared/Document/MRoot.cs
```diff
@@ -114,9 +114,9 @@ public FLSliceResult Encode()
 
         #region IDisposable
 
-        public void Dispose()
+        public override void Dispose()
         {
-            Context?.Dispose();
+            base.Dispose();
```
borrrden: There is no point in overriding a method just to call its base method.

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -54,6 +56,8 @@ internal sealed unsafe class FleeceMutableArray : MCollection, IList<object>, ID
 
         public bool IsEmpty => Native.FLArray_IsEmpty((FLArray*)_flArr);
 
+        //public bool IsMutated => Native.FLMutableArray_IsChanged(ar);
```
borrrden: Don't add in commented stuff

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -343,12 +355,14 @@ public override void FLEncode(FLEncoder* enc)
 
         #region IDisposable
 
-        public void Dispose()
+        public override void Dispose()
         {
-            if (_releaseRequired) {
+            base.Dispose();
+            if (_releaseRequired && _flArr != null) {
```
borrrden: It's ok to call release on null, it is a no-op

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -152,6 +155,15 @@ public void Set(int index, object val)
 
         #endregion
 
+        #region Internal Method(s)
+
+        internal void NewFLMutableArray()
```
borrrden: What is the point of having this as a separate method?  It seems dangerous because  new object won't be usable until this is called.

src/Couchbase.Lite.Shared/Serialization/FleeceMutableDictionary.cs
```diff
@@ -80,17 +79,18 @@ public void Clear()
...
 
             Mutate();
+
             _map.Clear();
-            foreach (var item in IterateDict()) {
-                _map[item.Key] = MValue.Empty;
-            }
 
-            Count = 0;
+            var list = IterateDict().AsParallel().ToList();
```
borrrden: `AsParallel()` ?  What is the goal of this?  The underlying logic is not parallel friendly

src/Couchbase.Lite.Shared/Serialization/FleeceMutableDictionary.cs
```diff
@@ -171,6 +159,12 @@ public void Set(string key, MValue val)
             }
         }
 
+        internal void NewFLMutableDict()
```
borrrden: Same comment as `NewFLMutableArray`

src/Couchbase.Lite.Shared/Serialization/FleeceMutableDictionary.cs
```diff
@@ -210,33 +209,59 @@ private FLDictIterator BeginIteration()
...
         {
             if (!IsMutated) {
-                if (_dict == null) {
-                    Native.FLEncoder_BeginDict(enc,0U);
+                if ((FLDict*)_dict == null) {
+                    Native.FLEncoder_BeginDict(enc, 0U);
                     Native.FLEncoder_EndDict(enc);
                 } else {
-                    Native.FLEncoder_WriteValue(enc, (FLValue*) _dict);
+                    Native.FLEncoder_WriteValue(enc, (FLValue*)(FLDict*)_dict);
```
borrrden: Is a direct cast to `FLValue *` not working here?  Because that seems like it would be fine.

src/Couchbase.Lite.Shared/Serialization/MCollection.cs
```diff
@@ -116,5 +117,14 @@ protected void Mutate()
         public abstract void FLEncode(FLEncoder* enc);
 
         #endregion
+
+        #region IDisposable
+
+        public virtual void Dispose()
+        {
+            Context?.Dispose();
```
borrrden: This makes me nervous because I don't remember who is responsible for managing this, but I don't think it is this class (what happens if two MCollections are made from the same context?)

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
Sandychuang8: Because a new FleeceMutableArray object will be created when an ArrayObject is created, but we don't want to call Native.FLMutableArray_New() and set _releaseRequired = true until ArrayObject() in constructed. And I want to do _threadSafety.DoLocked with Native.FLMutableArray_New()..

src/Couchbase.Lite.Shared/Serialization/MCollection.cs
```diff
@@ -122,7 +122,7 @@ protected void Mutate()
 
         public virtual void Dispose()
         {
-            Context?.Dispose();
+
```
borrrden: Why make this `IDisposable` just to have a blank method?  You had a better idea before with disposing the context here and then calling base if you needed to extend it (but I'm still not sure that any of these classes should be disposing the context.  They didn't before right?

Sandychuang8: MRoot did dispose Context before. MRoot extends MCollection.

borrrden: What benefit does it add to make `MCollection` an `IDisposable` object?

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
borrrden: You can do something like this inside of the `ArrayObject` constructor:

`_array = _threadSafety.DoLocked(() => new FLMutableArray())`

Sandychuang8: _array is FleeceMutableArray, and I already have internal readonly FleeceMutableArray _array = new FleeceMutableArray(); as variable in ArrayObject.

borrrden: So move the `new FleeceMutableArray()` to the constructor?  I don't see the benefit of having a separate method just to initialize an object

src/Couchbase.Lite.Shared/Serialization/FleeceMutableDictionary.cs
```diff
@@ -210,33 +209,59 @@ private FLDictIterator BeginIteration()
...
+        #endregion
+
+        #region IFLEncodable
+
         public override void FLEncode(FLEncoder* enc)
         {
             if (!IsMutated) {
-                if (_dict == null) {
-                    Native.FLEncoder_BeginDict(enc,0U);
+                if ((FLDict*)_dict == null) {
```
borrrden: Why is a cast needed here?

src/Couchbase.Lite.Shared/Serialization/FleeceMutableDictionary.cs
```diff
@@ -185,13 +179,18 @@ private bool Advance(ref FLDictIterator i)
...
 
+        private void EndIteration(FLDictIterator i)
+        {
+            Native.FLDictIterator_End(&i);
+        }
+
         private KeyValuePair<string, MValue> Get(FLDictIterator i)
         {
-            if (_dict == null || Native.FLDict_Count(_dict) == 0U) {
+            if ((FLDict*)_dict == null || Count == 0U) {
```
borrrden: Why is a cast needed here?

PR #1189, opened 2019-10-18
Enable Xamarin PR validation

Sandychuang8:


borrrden:
Weren't there other changes that were also needed in here to make this work correctly?

borrrden:
It looks like the validation machine needs to have Xcode updated.  Replay the job after that is done.

PR #1190, opened 2019-10-22
CBL-447 Add PendingDocumentIDs method and IsDocumentPending method

Sandychuang8:
CBL-447 Add PendingDocumentIDs method and IsDocumentPending method

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
+++ b/src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
@@ -58,6 +58,7 @@ namespace Couchbase.Lite
         internal const string NoAliasInJoin = "The default database must have an alias in order to use a JOIN statement (Make sure your data source uses the As() function).";
         internal const string InvalidQueryDBNull = "Invalid query: The database is null.";
         internal const string InvalidQueryMissingSelectOrFrom = "Invalid query: missing Select or From.";
+        internal const string PullOnlyPendingDocIDs = "Pending Document IDs are not supported on pull-only replicators.";
     }
 }
 
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -274,8 +274,9 @@ namespace Couchbase.Lite.Sync
         public IImmutableSet<string> GetPendingDocumentIDs()
         {
             if (!IsPushing()) {
-                WriteLog.To.Database.E(Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
-                throw new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+                CBDebug.LogAndThrow(WriteLog.To.Sync,
+                    new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs),
+                    Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs, true);
             }
 
             CreateReplicator();
@@ -296,12 +297,10 @@ namespace Couchbase.Lite.Sync
                 for (int i = 0; i < cnt; i++) {
                     var flv = Native.FLArray_Get(flarr, (uint)i);
                     result.Add(Native.FLValue_AsString(flv));
-                    Native.FLValue_Release(flv);
                 }
-
-                Native.FLValue_Release(flval);
             }
 
+            ids = null;
             return result.ToImmutableHashSet<string>();
         }
 
@@ -319,7 +318,8 @@ namespace Couchbase.Lite.Sync
             CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(documentID), documentID);
 
             if (!IsPushing()) {
-                CBDebug.LogAndThrow(WriteLog.To.Sync, new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs), 
+                CBDebug.LogAndThrow(WriteLog.To.Sync,
+                    new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs),
                     Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs, true);
             }
 
@@ -329,8 +329,7 @@ namespace Couchbase.Lite.Sync
             var isDocPending = Native.c4repl_isDocumentPending(_repl, documentID, &err);
 
             if (err.code > 0) {
-                WriteLog.To.Database.E(Tag, err.ToString());
-                throw CouchbaseException.Create(err);
+                CBDebug.LogAndThrow(WriteLog.To.Sync, CouchbaseException.Create(err), Tag, err.ToString(), true);
             }
 
             return isDocPending;
@@ -342,8 +341,7 @@ namespace Couchbase.Lite.Sync
 
         private bool IsPushing()
         {
-            var push = Config.ReplicatorType.HasFlag(ReplicatorType.Push);
-            return push;
+            return Config.ReplicatorType.HasFlag(ReplicatorType.Push);
         }
 
         private static C4ReplicatorMode Mkmode(bool active, bool continuous)
@@ -662,13 +660,10 @@ namespace Couchbase.Lite.Sync
         {
             _desc = ToString(); // Cache this; it may be called a lot when logging
 
-            var err = new C4Error();
             var status = default(C4ReplicatorStatus);
             _stopping = false;
 
-            if (_repl == null) {
-                err = CreateReplicator();
-            }
+            var err = CreateReplicator();
 
             _databaseThreadSafety.DoLocked(() =>
             {
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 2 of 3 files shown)


src/Couchbase.Lite.Shared/API/Error/CouchbaseLiteErrorMessage.cs
```diff
@@ -58,6 +58,7 @@ internal static partial class CouchbaseLiteErrorMessage
         internal const string NoAliasInJoin = "The default database must have an alias in order to use a JOIN statement (Make sure your data source uses the As() function).";
         internal const string InvalidQueryDBNull = "Invalid query: The database is null.";
         internal const string InvalidQueryMissingSelectOrFrom = "Invalid query: missing Select or From.";
+        internal const string PullOnlyPendingDocIDs = "Pending Document IDs are not supported on pull-only replicators.";
```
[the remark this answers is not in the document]
Sandychuang8: yes, it's there: https://github.com/couchbaselabs/couchbase-mobile-tools/blob/master/gen_errors/cbl-errors.json

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
```diff
@@ -45,11 +45,32 @@ internal unsafe static partial class Native
         [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
         public static extern C4ReplicatorStatus c4repl_getStatus(C4Replicator* repl);
 
+        public static byte[] c4repl_getPendingDocIDs(C4Replicator* repl, C4Error* outErr)
+        {
+            using(var retVal = NativeRaw.c4repl_getPendingDocIDs(repl, outErr)) {
+                return ((FLSlice)retVal).ToArrayFast();
```
jayahariv: what if `retVal` is nil from lite-core?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,83 @@ public void Stop()
...
+        {
+            CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(documentID), documentID);
+
+            if (IsNotPushing()) {
+                WriteLog.To.Database.E(Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+                throw new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+            }
+
+            C4Error err;
+            var isDocPending = Native.c4repl_isDocumentPending(_repl, documentID, &err);
```
jayahariv: should we handle the `err`? make sure, there is no error?

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
Sandychuang8: This class is actually C# binding class... Auto generated.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: oh, you are right, I forgot to handle error here.

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
Sandychuang8: https://github.com/couchbase/couchbase-lite-core/commit/82df7b625b23b0540bad52e8667e304b78b83f10#diff-97c2a0f6c7da83884c12f2f63f627122R244 <--- possibly from here

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,83 @@ public void Stop()
...
+
         #endregion
 
         #region Private Methods
 
+        private bool IsNotPushing()
+        {
+            var push = Config.ReplicatorType.HasFlag(ReplicatorType.Push);
+            var pushnpull = Config.ReplicatorType.HasFlag(ReplicatorType.PushAndPull);
+            return !push && !pushnpull;
```
jayahariv: (optional) making a positive check will be more straight forward
```
private bool isPushing() {
    // .....
    return push || pushNPull
}

// when using 
if (~isPushing()) {

}
```

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,89 @@ public void Stop()
...
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        [NotNull]
+        public IImmutableSet<string> GetPendingDocumentIDs()
+        {
+            if (!IsPushing()) {
+                WriteLog.To.Database.E(Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+                throw new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+            }
+
+            var result = new HashSet<string>();
```
jayahariv: every time we are creating a new hashset, add all ids, and return back, but did we discussed about caching results?

@borrrden

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
```diff
@@ -45,11 +45,32 @@ internal unsafe static partial class Native
...
+            }
+        }
+
+        public static bool c4repl_isDocumentPending(C4Replicator* repl, string docID, C4Error* outErr)
+        {
+            using(var docID_ = new C4String(docID)) {
+                return NativeRaw.c4repl_isDocumentPending(repl, docID_.AsFLSlice(), outErr);
+            }
+        }
+
```
jayahariv: extra line?

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
```diff
@@ -45,11 +45,32 @@ internal unsafe static partial class Native
...
 
     internal unsafe static partial class NativeRaw
     {
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        public static extern FLSliceResult c4repl_getPendingDocIDs(C4Replicator* repl, C4Error* outErr);
+
+        [DllImport(Constants.DllName, CallingConvention = CallingConvention.Cdecl)]
+        [return: MarshalAs(UnmanagedType.U1)]
+        public static extern bool c4repl_isDocumentPending(C4Replicator* repl, FLSlice docID, C4Error* outErr);
+
```
jayahariv: extra line??

jayahariv: since auto generated, i am resolving this.

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
jayahariv: since auto generated, i am resolving this.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: Yes the results should be cached against the value of the local sequence of the pusher.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,97 @@ public void Stop()
...
+        /// </item>
+        /// </summary>
+        /// <returns>An immutable set of strings, each of which is a document ID</returns>
+        /// <exception cref="CouchbaseLiteException">Thrown if no push replication</exception>
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        [NotNull]
+        public IImmutableSet<string> GetPendingDocumentIDs()
+        {
+            if (!IsPushing()) {
+                WriteLog.To.Database.E(Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
```
borrrden: There is a helper function to both log and throw an exception that you might want to use here.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,97 @@ public void Stop()
...
+                throw new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+            }
+
+            if (_repl == null) {
+                StartInternal();
```
borrrden: This is not part of the technical spec, why is this starting the replicator?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: Use the .NET convention that is used everywhere else for throwing errors like this.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: Actually mainly creating a replicator.... because test will fail when replicator is null..

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,97 @@ public void Stop()
...
+
         #endregion
 
         #region Private Methods
 
+        private bool IsPushing()
+        {
+            var push = Config.ReplicatorType.HasFlag(ReplicatorType.Push);
+            var pushnpull = Config.ReplicatorType.HasFlag(ReplicatorType.PushAndPull);
+            return push || pushnpull;
```
borrrden: `pushnpull` here is redundant.  The first line will return true for both cases (that's the meaning of "has flag")

src/LiteCore/src/LiteCore.Shared/Interop/C4Replicator_native.cs
borrrden: By the way @jayahariv `FLSlice` and `FLSliceResult` cannot be null since they are value types.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: What test?  If the replicator is not started then this should simply return no documents.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +266,97 @@ public void Stop()
...
+            }
+
+            if (ids != null) {
+                var flval = Native.FLValue_FromData(ids, FLTrust.Trusted);
+                var flarr = Native.FLValue_AsArray(flval);
+                var cnt = (int)Native.FLArray_Count(flarr);
+
+                for (int i = 0; i < cnt; i++) {
+                    var flv = Native.FLArray_Get(flarr, (uint)i);
+                    result.Add(Native.FLValue_AsString(flv));
```
jayahariv: also, this might be a `FLString`(Native.FLValue_AsString(flv)), which might need to convert it to `string`

Sandychuang8: Native.FLValue_AsString(flv) converts array item to string.

jayahariv: oh is it, sorry about that.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -229,11 +230,6 @@ public void Start()
                     throw new ObjectDisposedException(CouchbaseLiteErrorMessage.ReplicatorDisposed);
                 }
 
-                if (_repl != null) {
-                    WriteLog.To.Sync.W(Tag, $"{this} has already started");
-                    return;
-                }
```
borrrden: There will still need to be a way to avoid a double start, though!

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: Apply `LogAndThrow` everywhere that you can.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +261,91 @@ public void Stop()
...
+
+            CreateReplicator();
+
+            C4Error err;
+            var isDocPending = Native.c4repl_isDocumentPending(_repl, documentID, &err);
+
+            if (err.code > 0) {
+                WriteLog.To.Database.E(Tag, err.ToString());
+                throw CouchbaseException.Create(err);
+            }
```
borrrden: This is a common pattern, and it has a helper function already.  See `LiteCoreBridge.Check`

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +261,91 @@ public void Stop()
...
+        }
+
         #endregion
 
         #region Private Methods
 
+        private bool IsPushing()
+        {
+            var push = Config.ReplicatorType.HasFlag(ReplicatorType.Push);
+            return push;
```
borrrden: This is basically one line, not two.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +261,91 @@ public void Stop()
...
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        [NotNull]
+        public IImmutableSet<string> GetPendingDocumentIDs()
+        {
+            if (!IsPushing()) {
+                WriteLog.To.Database.E(Tag, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+                throw new CouchbaseLiteException(C4ErrorCode.Unsupported, CouchbaseLiteErrorMessage.PullOnlyPendingDocIDs);
+            }
+
+            CreateReplicator();
```
borrrden: Bad!  Don't unilaterally create the replicator here.  What if it already exists?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -265,10 +261,91 @@ public void Stop()
...
+
+            if (ids != null) {
+                var flval = Native.FLValue_FromData(ids, FLTrust.Trusted);
+                var flarr = Native.FLValue_AsArray(flval);
+                var cnt = (int)Native.FLArray_Count(flarr);
+
+                for (int i = 0; i < cnt; i++) {
+                    var flv = Native.FLArray_Get(flarr, (uint)i);
+                    result.Add(Native.FLValue_AsString(flv));
+                    Native.FLValue_Release(flv);
```
borrrden: `FLValue_Release` is only for values that you create or retain.  There is none of that here.  You just need to free `ids` at the end of this.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -296,12 +297,10 @@ public IImmutableSet<string> GetPendingDocumentIDs()
                 for (int i = 0; i < cnt; i++) {
                     var flv = Native.FLArray_Get(flarr, (uint)i);
                     result.Add(Native.FLValue_AsString(flv));
-                    Native.FLValue_Release(flv);
                 }
-
-                Native.FLValue_Release(flval);
             }
 
+            ids = null;
```
borrrden: This is a memory leak, you need to release it explicitly.

borrrden:
I think a lot of this stuff will come out with tests so add them to this PR.

PR #1191, opened 2019-11-04
CBL-510 Check both remote and local docs are deleted and handle the case.

Sandychuang8:
CBL-510 Check both remote and local docs are deleted and handle the case. Add unit test to verify.


borrrden:
Add me back when the validation is passing.

PR #1192, opened 2019-11-05
CBL-486 Wait for CCR tasks completed before replication status idle callback

Sandychuang8:
CBL-486 Wait for CCR tasks completed before replication status idle callback


borrrden:
I'm wondering if there are any tests for this.  @jayahariv Are there any tests on iOS?

jayahariv:
no, we haven't added any, but its a good idea to include one.

PR #1194, opened 2019-11-14
CBL-499 fix RevisionID not null inside push/pull filter

Sandychuang8:
CBL-499 fix RevisionID not null inside push/pull filter


Sandychuang8:
9th PR Validation run all platforms pass. But not sure it points to the 7th run....

PR #1201, opened 2020-01-13
Feature/cbl 628 unit test conflic flag clear after resolve

Sandychuang8:
CBL-628 unit test conflict flag clear after resolve

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -1666,6 +1666,40 @@ namespace Test
             TestConflictResolverExceptionThrown(blobFromOtherDbResolver, false, true);
         }
 
+        //CBL-623: Revision flags get cleared while saving resolved document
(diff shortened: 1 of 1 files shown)


src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1750,39 +1784,62 @@ private void TestConflictResolverWins(bool returnRemoteDoc)
             }
         }
 
-        private void CreateReplicationConflict(string id)
+        private void CreateReplicationConflict(string id, bool checkFlags = false)
```
borrrden: This is not good for long term maintenance.  I guarantee everyone will forgot what "checkFlags" means here.  Maybe pass in a nullable set of expected flags (for mercury time sake it can be this way though).

Sandychuang8: I will cherry pick this into master and replace this with your suggestion.

PR #1202, opened 2020-02-10
Add release Blob's ContentStream reminder after it's usage

Sandychuang8:
Add release Blob's ContentStream reminder after it's usage

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Document/Blob.cs
+++ b/src/Couchbase.Lite.Shared/API/Document/Blob.cs
@@ -127,6 +127,9 @@ namespace Couchbase.Lite
         /// <summary>
         /// Gets the contents of the blob as a <see cref="Stream"/>
         /// </summary>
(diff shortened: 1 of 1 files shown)


Sandychuang8:
Please ignore the failure of PR Validation in this PR.
PR validation will not pass due to unit tests are not all fixed in master branch with the latest LiteCore updates.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -127,6 +127,9 @@ public byte[] Content
         /// <summary>
         /// Gets the contents of the blob as a <see cref="Stream"/>
         /// </summary>
+        /// <remarks>
+        /// Please release all resources of this Stream object soon after usage.
```
borrrden: "The caller is responsible for disposing the Stream when finished with it"

PR #1203, opened 2020-02-17
Feature/use fl slot set methods

Sandychuang8:
CBL-650 
Part 1: Fix tests in ArrayTest and CSharpTest mainly by updating current way of value setting in FLMutableArray and converting FLValue from FLSliceResult using FLDoc API
Part 2: Fix Predictive query tests with AggregateModel (https://github.com/couchbase/couchbase-lite-net/pull/1204)
Part 3: Replication tests (https://github.com/couchbase/couchbase-lite-net/pull/1205)


src/LiteCore/parse/templates_c4/C4ReplicatorParameters.cs
```diff
@@ -11,13 +11,4 @@ internal unsafe partial struct C4ReplicatorParameters
         public void* callbackContext;
         public C4SocketFactory* socketFactory;
         private byte _dontStart;
```
borrrden: If `dontStart` is removed `_dontStart` should go with it

src/LiteCore/parse/templates_c4/C4SocketFactory.cs
```diff
@@ -1,4 +1,4 @@
-    internal unsafe struct C4SocketFactory
+    internal unsafe partial struct C4SocketFactory
```
borrrden: What is the purpose of making this partial (and the others that were made partial)?

Sandychuang8: I thought they are generated by binding script?

borrrden: This one is a template, and the binding script configuration was also modified as part of this PR.  I'm just wondering if there is a reason to change them.

Sandychuang8: There is no reason to change the script anymore. Let me try to revert them all.

PR #1205, opened 2020-02-28
Feature/replicator dev

Sandychuang8:
CBL-776
CBL-447 Pending Document Ids (https://github.com/couchbase/couchbase-lite-net/pull/1206)

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -581,8 +581,7 @@ namespace Couchbase.Lite.Sync
             StartInternal();
         }
 
-        // Must be called from within the ThreadSafety
-        private void StartInternal()
+        private C4Error SetupC4Replicator()
         {
             _desc = ToString(); // Cache this; it may be called a lot when logging
 
@@ -632,14 +631,12 @@ namespace Couchbase.Lite.Sync
             // Clear the reset flag, it is a one-time thing
             options.Reset = false;
 
-            if(Config.PushFilter!=null)
+            if (Config.PushFilter != null)
                 _nativeParams.PushFilter = PushFilterCallback;
             if (Config.PullFilter != null)
                 _nativeParams.PullFilter = PullValidateCallback;
 
-            var err = new C4Error();
-            var status = default(C4ReplicatorStatus);
-            
+            C4Error err = new C4Error();
             _databaseThreadSafety.DoLocked(() =>
             {
                 C4Error localErr = new C4Error();
@@ -653,6 +650,23 @@ namespace Couchbase.Lite.Sync
                 }
 
                 err = localErr;
+            });
+
+            scheme.Dispose();
+            path.Dispose();
+            host.Dispose();
+
+            return err;
+        }
+
+        // Must be called from within the ThreadSafety
+        private void StartInternal()
+        {
+            var err = SetupC4Replicator();
+            var status = default(C4ReplicatorStatus);
+            
+            _databaseThreadSafety.DoLocked(() =>
+            {
                 if (_repl != null && _stopping) {
                     _stopping = false;
                     Native.c4repl_start(_repl);
@@ -666,10 +680,6 @@ namespace Couchbase.Lite.Sync
                     };
                 }
             });
-
-            scheme.Dispose();
-            path.Dispose();
-            host.Dispose();
 
             UpdateStateProperties(status);
             DispatchQueue.DispatchSync(() => StatusChangedCallback(status));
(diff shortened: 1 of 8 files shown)


borrrden:
PR validation is never going to work for this because it's not against a branch that also exists in EE.  To fix it you can temporarily make a branch in EE called feature/CBL-774

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -197,9 +192,12 @@ public ListenerToken AddDocumentReplicationListener([NotNull]EventHandler<Docume
...
         {
-            _statusChanged.Remove(token);
-            if(_documentEndedUpdate.Remove(token)==0)
-                Config.Options.ProgressLevel = ReplicatorProgressLevel.Overall;
+            DispatchQueue.DispatchSync(() =>
+            {
+                _statusChanged.Remove(token);
+                if (_documentEndedUpdate.Remove(token) == 0)
+                    Config.Options.ProgressLevel = ReplicatorProgressLevel.Overall;
+            });
```
borrrden, line 200: What problem did this fix?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -223,44 +221,51 @@ public void ResetCheckpoint()
         /// </summary>
         public void Start()
         {
+            if (_disposed) {
+                throw new ObjectDisposedException(CouchbaseLiteErrorMessage.ReplicatorDisposed);
+            }
```
borrrden, line 226: Why was this moved outside?  The disposed check could now be incorrect.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -223,44 +221,51 @@ public void ResetCheckpoint()
...
         public void Stop()
         {
             DispatchQueue.DispatchSync(() =>
             {
-                if (_stopping) {
-                    return;
-                }
-                
-                _stopping = true;
-                _reachability?.Stop();
```
borrrden, line 255: Why was this removed?  The reason it is here is so that the reachability callback does not interfere with the stopping process.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -398,16 +393,6 @@ private void WaitPendingConflictTasks(C4ReplicatorStatus status)
             }
         }
 
-        private void ClearRepl()
-        {
-            DispatchQueue.DispatchSync(() =>
-            {
-                Native.c4repl_free(_repl);
```
borrrden, line 405: Where is this freed now?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -223,44 +221,51 @@ public void ResetCheckpoint()
...
+                    status = new C4ReplicatorStatus {
+                        error = err,
+                        level = C4ReplicatorActivityLevel.Stopped,
+                        progress = new C4Progress()
+                    };
                 }
-
-                WriteLog.To.Sync.I(Tag, $"{this}: Starting");
-                _retryCount = 0;
-                StartInternal();
```
borrrden, line 239: I don't see where `StartInternal` gets called now....does it ever?

borrrden: Nevermind, I see that was merged into `SetupC4Replicator`

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: Native.c4repl_free(_repl); is now called in Dispose/Finalizing.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: StopReachabilityObserver() is called when replicator is in idle or busy and stopped states. Is `c4repl_setHostReachable` taking care Reachability Observing on the LiteCore side?

borrrden: All that method does is react to the information that is provided.  The platform still needs to monitor the actual condition and inform LiteCore.  If stop is called though, the monitoring should immediately stop.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: Oh ok, that was already there!

PR #1206, opened 2020-03-02
Feature/pending doc ids

Sandychuang8:
CBL-447 - Add get pending doc ids and check is document pending methods. - Add unit tests.


borrrden:
PR validation is not going to work for this either for the same reason as #1205

PR #1209, opened 2020-04-10
Feature/cbl 849

Sandychuang8:
CBL-849 Returning WebProxy null when Wwhen proxy server list is null in WINHTTP_PROXY_INFO.


borrrden:
Failed to build on Linux, but this class is not used on Linux anyway so I am not concerned.  Need to create an action item to update the validation project fetch logic though.

PR #1210, opened 2020-04-16
CBL-774 Enhance Database Close and Delete

Sandychuang8:
CBL-774 Enhance Database Close and Delete
Only unit test on live queries, replications tests will be added after replicator are updated base on LiteCore updates.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -24,7 +24,6 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Diagnostics.CodeAnalysis;
 
 using Couchbase.Lite.DI;
 using Couchbase.Lite.Internal.Doc;
@@ -42,8 +41,6 @@ using LiteCore.Interop;
 using LiteCore.Util;
 
 using Newtonsoft.Json;
-
-using JetBrains.Annotations;
 
 using NotNull = JetBrains.Annotations.NotNullAttribute;
 using ItemNotNull = JetBrains.Annotations.ItemNotNullAttribute;
@@ -126,6 +123,8 @@ namespace Couchbase.Lite
         private C4DatabaseObserver* _obs;
         private GCHandle _obsContext;
         private C4Database* _c4db;
+        private bool _isClosing;
+        private ManualResetEvent _closeCondition = new ManualResetEvent(false);
 
         #endregion
 
@@ -208,9 +207,6 @@ namespace Couchbase.Lite
             }
         }
 
-        [@NotNull]
-        internal IDictionary<Uri, Replicator> Replications { get; } = new Dictionary<Uri, Replicator>();
-
         internal FLEncoder* SharedEncoder
         {
             get {
@@ -228,7 +224,33 @@ namespace Couchbase.Lite
         [@NotNull]
         internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
 
-        private bool IsShell { get; }
+        internal bool IsClosedLocked
+        {
+            get {
+                return ThreadSafety.DoLocked(() => {
+                    return IsClosed;
+                });
+            }
+        }
+
+        private bool IsShell { get; } //this object is borrowing the C4Database from somewhere else, so don't free C4Database at the end if isshell
+
+        // Must be called inside self lock
+        private bool IsClosed
+        {
+            get {
+                return _c4db == null;
+            }
+        }
+
+        private bool IsReadyToClose
+        {
+            get {
+                return ThreadSafety.DoLocked(() => {
+                    return ActiveReplications.Count == 0 && ActiveLiveQueries.Count == 0;
+                });
+            }
+        }
 
         #endregion
 
@@ -880,6 +902,24 @@ namespace Couchbase.Lite
 
         #region Internal Methods
 
+        internal void AddActiveLiveQuery(XQuery query)
+        {
+            ThreadSafety.DoLocked(() => {
+                CheckOpenAndNotClosing();
+                ActiveLiveQueries.Add(query);
+            });
+        }
+
+        internal void RemoveActiveLiveQuery(XQuery query)
+        {
+            ThreadSafety.DoLocked(() => {
+                ActiveLiveQueries.Remove(query);
+                if (ActiveLiveQueries.Count == 0) {
+                    _closeCondition.Set();
+                }
+            });
+        }
+
         internal void ResolveConflict([@NotNull]string docID, [@CanBeNull]IConflictResolver conflictResolver)
         {
             Debug.Assert(docID != null);
@@ -1026,29 +1066,27 @@ namespace Couchbase.Lite
 
         private void CheckOpen()
         {
-            if(_c4db == null) {
+            if(IsClosed) {
                 throw new InvalidOperationException(CouchbaseLiteErrorMessage.DBClosed);
             }
         }
 
         private void Dispose(bool disposing)
         {
-            if (_c4db == null) {
+            if (IsClosed) {
                 return;
             }
 
             if (disposing) {
-                if (_obs != null) {
-                    Native.c4dbobs_free(_obs);
-                    _obsContext.Free();
-                }
-
+                //TODO _docObs might need to be refactored into an IDisposable class
                 foreach (var obs in _docObs) {
                     Native.c4docobs_free((C4DocumentObserver *)obs.Value.Item1);
                     obs.Value.Item2.Free();
                 }
 
                 _docObs.Clear();
+                //end of TODO comments
+
                 if (_unsavedDocuments.Count > 0) {
                     WriteLog.To.Database.W(Tag,
                         $"Closing database with {_unsavedDocuments.Count} such as {_unsavedDocuments.Any()}");
@@ -1057,13 +1095,17 @@ namespace Couchbase.Lite
                 _unsavedDocuments.Clear();
             }
 
+            if (_obs != null) {
+                Native.c4dbobs_free(_obs);
+                _obsContext.Free();
+            }
+
             WriteLog.To.Database.I(Tag, $"Closing database at path {Native.c4db_getPath(_c4db)}");
             if (!IsShell) {
                 LiteCoreBridge.Check(err => Native.c4db_close(_c4db, err));
             }
 
-            Native.c4db_release(_c4db);
-            _c4db = null;
+            FreeC4Db();
         }
 
         [@CanBeNull]
@@ -1131,7 +1173,7 @@ namespace Couchbase.Lite
         {
 			ThreadSafety.DoLocked(() =>
 			{
-				if (_obs == null || _c4db == null) {
+				if (_obs == null || IsClosed) {
 					return;
 				}
 
@@ -1167,7 +1209,7 @@ namespace Couchbase.Lite
             DocumentChangedEventArgs change = null;
             ThreadSafety.DoLocked(() =>
             {
-                if (_c4db == null || !_docObs.ContainsKey(documentID) || Native.c4db_isInTransaction(_c4db)) {
+                if (IsClosed || !_docObs.ContainsKey(documentID) || Native.c4db_isInTransaction(_c4db)) {
                     return;
                 }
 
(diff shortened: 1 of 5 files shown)


src/Couchbase.Lite.Tests.Shared/QueryTest.cs
```diff
@@ -1577,11 +1577,11 @@ public void TestLiveQueryBlocksClose()
                 }
 
                 doc1Listener.WaitForResult(TimeSpan.FromSeconds(20));
-                otherDb.Invoking(d => d.Dispose())
-                    .Should().Throw<CouchbaseLiteException>("because the live query is still active");
+                //otherDb.Invoking(d => d.Dispose())
+                //    .Should().Throw<CouchbaseLiteException>("because the live query is still active");
```
Sandychuang8, line 1581: I assume this exception will no longer throw because live queries and replications should all stopped before database closing.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -540,14 +563,12 @@ public void CreateIndex([@NotNull]string name, [@NotNull]IIndex index)
         /// <exception cref="InvalidOperationException">Thrown if this method is called after the database is closed</exception>
         public void Delete()
         {
-            StopExpirePurgeTimer();
```
borrrden, line 543: Looks like this was mistakenly removed?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1552,58 @@ public void Dispose()
...
+                }
+
+                //TODO _docObs might need to be refactored into an IDisposable class
+                foreach (var obs in _docObs) {
+                    Native.c4docobs_free((C4DocumentObserver*)obs.Value.Item1);
+                    obs.Value.Item2.Free();
+                }
+
+                _docObs.Clear();
+                //end of TODO comments
```
borrrden, line 1587: The above logic is duplicated with `Dispose(bool)`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -228,7 +225,33 @@ public string Path
         [@NotNull]
         internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
 
-        private bool IsShell { get; }
+        internal bool IsClosedLocked
```
borrrden, line 228: This is never used except in tests?  What is it for?

src/Couchbase.Lite.Tests.Shared/QueryTest.cs
```diff
@@ -1577,11 +1577,11 @@ public void TestLiveQueryBlocksClose()
...
                 doc1Listener.WaitForResult(TimeSpan.FromSeconds(20));
-                otherDb.Invoking(d => d.Dispose())
-                    .Should().Throw<CouchbaseLiteException>("because the live query is still active");
+                //otherDb.Invoking(d => d.Dispose())
+                //    .Should().Throw<CouchbaseLiteException>("because the live query is still active");
             } finally {
-                query.RemoveChangeListener(token);
-                query.Dispose();
+                //query.RemoveChangeListener(token);
+                //query.Dispose();
```
borrrden, line 1584: Remove this test if it is no longer applicable

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Because it's handled in Close/Dispose method.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: yes, its purpose of existence is for testing. iOS has it in their test...

borrrden: Why is it needed just for a test?

borrrden: FYI I talked to @pasin and he will remove it from iOS

src/Couchbase.Lite.Tests.Shared/QueryTest.cs
pasin: https://issues.couchbase.com/browse/CBL-892

src/Couchbase.Lite.Shared/API/Database/Database.cs
pasin: https://issues.couchbase.com/browse/CBL-892

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1552,56 @@ public void Dispose()
...
+                    return;
+                }
+
+                if (!_isClosing) {
+                    _isClosing = true;
+                }
+            });
+
+            _closeCondition.Reset();
+            foreach (var q in ActiveLiveQueries.ToList()) {
```
pasin, line 1567: Is ActiveLiveQueries.ToList() thread safe?

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: strange, I do see iOS is using `Database.isClosedLocked()` in their unit tests.
for ex: https://github.com/couchbase/couchbase-lite-ios/commit/f83efb4b6d01c525982b9f38acb953b2d34bcd03#diff-124d009e59a6f172bad43ef7c3856c78R1795

pasin: It's optional. So please decide whether to have it or not based on your implementation and test. I will let @jayahariv double check this on iOS.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -179,11 +180,11 @@ public string Path
 
         [@NotNull]
         [@ItemNotNull]
-        internal ICollection<XQuery> ActiveLiveQueries { get; } = new HashSet<XQuery>();
+        internal ConcurrentDictionary<XQuery, int> ActiveLiveQueries { get; } = new ConcurrentDictionary<XQuery, int>();
```
borrrden, line 183: What is the point of making these concurrent if you wrap them all in thread safety anyway?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1571,56 @@ public void Dispose()
...
+                }
+            });
+
+            _closeCondition.Reset();
+            foreach (var q in ActiveLiveQueries) {
+                q.Key.Stop();
+            }
+
+            while (!IsReadyToClose) {
+                _closeCondition.WaitOne();
```
borrrden, line 1591: This loop is not valid.  It will wait once, and then have a tight loop because the condition remains set forever.  Consider using `AutoResetEvent` instead so that each time you wait, it will automatically reset.  Furthermore, it would be good to dispose it in the database dispose method.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1571,56 @@ public void Dispose()
...
+                    return;
+                }
+
+                if (!_isClosing) {
+                    _isClosing = true;
+                }
+            });
+
+            _closeCondition.Reset();
+            foreach (var q in ActiveLiveQueries) {
```
borrrden, line 1586: What about stopping replicators?

src/Couchbase.Lite.Tests.Shared/DatabaseTest.cs
```diff
@@ -1259,6 +1259,18 @@ public void TestGetDocFromDeletedDB()
             }
         }
 
+        [Fact]
+        public void TestCloseWithActiveLiveQueries()
+        {
+            WithActiveLiveQueries(true);
+        }
```
borrrden, line 1266: Active replicator tests?

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: My guess: To allow a non-locked loop in the close method?

src/Couchbase.Lite.Tests.Shared/DatabaseTest.cs
Sandychuang8: They are in the other PR (Replicator dev) which has all the replication updates.

borrrden: Alright, so CBL-774 can be closed after both of those are merged

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: In the replicator dev PR as well (https://github.com/couchbase/couchbase-lite-net/pull/1205/files#diff-795d11b3c81c750fb0cd00714df85829R1720)

src/Couchbase.Lite.Tests.Shared/DatabaseTest.cs
Sandychuang8: if 774 is approved, 774 can be merged into master, and I will rebase master into replicator dev branch, and replicator dev can point to master so replicator dev pr will show only replicator updates. cleaner this way..

pasin: I didn't go look at the other PR. Why does making the replication change in a separate PR? This makes this PR incomplete and is hard to know if anything missing.

Sandychuang8: replication dev PR is a big one, but if you like, we can merge this pr into the replication dev pr. but replication pr will not only contain database close/delete, it will also contain lots of replication updates..

pasin: Then don't merge.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: yes, I believe that's right.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1586,42 @@ public void Dispose()
...
+                }
+            });
+
+            foreach (var q in ActiveLiveQueries) {
+                q.Key.Stop();
+            }
+
+            while (!IsReadyToClose) {
+                _closeCondition.WaitOne();
+                _closeCondition.Reset();
```
borrrden, line 1606: The point of `AutoResetEvent` is that you don't have to call reset each time, it does it automatically.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1498,11 +1586,42 @@ public void Dispose()
...
+            }
+
+            ThreadSafety.DoLocked(() => {
+                if (IsClosed) {
+                    return;
+                }
+
+                ClearUnsavedDocsAndFreeDocObservers();
+                FreeC4DbObserver();
```
borrrden, line 1615: What is the difference between here and `Dispose(bool)` which appears to largely do the same stuff?  This will cause confusion and problems later if you have to maintain two places that are doing the same things.

src/Couchbase.Lite.Shared/Query/XQuery.cs
```diff
@@ -58,7 +58,7 @@ internal class XQuery : IQuery
         private DateTime _lastUpdatedAt;
         private int _observingCount = 0;
         [NotNull]private Parameters _queryParameters = new Parameters();
-        private AtomicBool _willUpdate = false;
+        private AtomicBool _willUpdate = false, _updating = false, _stopping = false;
```
borrrden, line 61: FYI there is no reason to have these be atomic bools.  The reason to have an atomic bool is to be able to set and check the previous value as one atomic step (i.e. `if foo.set(value)`) so these two added ones could just be normal bools

src/Couchbase.Lite.Tests.Shared/QueryTest.cs
```diff
@@ -1577,8 +1577,6 @@ public void TestLiveQueryBlocksClose()
                 }
 
                 doc1Listener.WaitForResult(TimeSpan.FromSeconds(20));
-                otherDb.Invoking(d => d.Dispose())
```
borrrden, line 1580: There is no reason to keep this test around as it tests behavior that is now incorrect.  Keeping it around will only confuse people who look at it later because the name suggests that live query *should* be blocking the close when actually the test no longer tests that with this line removed.

PR #1214, opened 2020-05-07
Add pending doc ids feature and unit tests

Sandychuang8:
CBL-450 CBL-451 CBL-452 CBL-453

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
@@ -75,6 +76,7 @@ namespace Couchbase.Lite.Sync
         private IReachability _reachability;
         private C4Replicator* _repl;
         private ConcurrentDictionary<Task, int> _conflictTasks = new ConcurrentDictionary<Task, int>();
+        private IImmutableSet<string> _pendingDocIds;
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -269,10 +271,104 @@ public void Stop()
...
+                    CBDebug.LogAndThrow(WriteLog.To.Sync, CouchbaseException.Create(errSetupRepl), Tag, errSetupRepl.ToString(), true);
+                }
+            });
+
+            C4Error err = new C4Error();
+            byte[] pendingDocIds = null;
+            pendingDocIds = Native.c4repl_getPendingDocIDs(_repl, &err);
+
+            if (err.code > 0) {
+                CBDebug.LogAndThrow(WriteLog.To.Sync, CouchbaseException.Create(err), Tag, err.ToString(), true);
```
borrrden, line 305: Shouldn't this be an exception?

borrrden: Duh....I read it wrong....but use `LiteCoreBridge` for this purpose.  It saves lines.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -269,10 +271,104 @@ public void Stop()
...
+                var errSetupRepl = SetupC4Replicator();
+                if (errSetupRepl.code > 0) {
+                    CBDebug.LogAndThrow(WriteLog.To.Sync, CouchbaseException.Create(errSetupRepl), Tag, errSetupRepl.ToString(), true);
+                }
+            });
+
+            C4Error err = new C4Error();
+            isDocPending = Native.c4repl_isDocumentPending(_repl, documentID, &err);
+            if (err.code > 0) {
+                CBDebug.LogAndThrow(WriteLog.To.Sync, CouchbaseException.Create(err), Tag, err.ToString(), true);
```
borrrden, line 357: This is what `LiteCoreBridge` is for (and above as well)

Sandychuang8: Looks like LiteCoreBridge.Check doesn't get me what I need.. In the pending doc ids, I need retrieve byte[] as pending doc ids (converted from FLSlice in C4Replicator_native) and bool as is document pending..

borrrden: It might not be set up to handle `byte[]` but it can be modified to do so if desired.  It already handles bools though via this delegate signature https://github.com/couchbase/couchbase-lite-net/blob/26e597cc00a9245a3ace9b7ff39da1994b84443e/src/LiteCore/src/LiteCore.Shared/Interop/NativeHandler.cs#L30-L34

PR #1235, opened 2020-08-18
Don't send the client certificate data to LiteCore

borrrden:
Instead just send a token for it to claim later.  This avoids logging sensitive information as well because all that will be logged is a memory address pointing to a handle to a .NET object.


borrrden:
PR validation failed but looks unrelated to this change

PR #1238, opened 2020-08-21
CBL-396 Manage HTTP cookies

Sandychuang8:
CBL-396 Manage HTTP cookies


borrrden (approved):
The changes look ok from a reading of things but needs to be tested in some fashion

PR #1241, opened 2020-08-26
CBL-1217 Check database is open in the beginning of pending doc id API

Sandychuang8:
CBL-1217 Check Database is open before proceeding getting pending doc ids and check is doc pending. Add unit tests to verify.


borrrden (approved):
Assuming that the exception type is equivalent to the other platform errors (i.e do they use CouchbaseLite errors?) it's fine

Sandychuang8:
Yes, iOS is using NSInternalInconsistencyException.

PR #1242, opened 2020-08-28
Use SecurityString in password securely. 

Sandychuang8:
Only convert the SecurityString to string in fleece encoding and password into NetworkCredential.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/BasicAuthenticator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/BasicAuthenticator.cs
@@ -50,7 +50,7 @@ namespace Couchbase.Lite.Sync
         /// <summary>
         /// [DEPRECATED] Gets the password that this object holds
         /// </summary>
-        /// [Obsolete("This property is deprecated, please use byte[] PasswordData instead.")]
+        /// [Obsolete("This property is deprecated, please use SecureString PasswordSecureString instead.")]
         [NotNull]
         public string Password { get; }
 
@@ -58,7 +58,7 @@ namespace Couchbase.Lite.Sync
         /// Gets the password that this object holds
         /// </summary>
         [NotNull]
-        internal SecureString PasswordSecureString { get; }
+        public SecureString PasswordSecureString { get; }
 
         #endregion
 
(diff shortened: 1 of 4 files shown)


src/LiteCore/src/LiteCore.Shared/Interop/Fleece.cs
```diff
@@ -549,6 +551,9 @@ public static void FLEncode(this object obj, FLEncoder* enc)
                 case Blob b:
                     b.FLEncode(enc);
                     break;
+                case SecureString ss:
+                    new NetworkCredential(string.Empty, ss).Password.FLEncode(enc);
```
Sandychuang8, line 555: The NetworkCredential class internally always stores the password as a SecureString:
https://referencesource.microsoft.com/#System/net/System/Net/NetworkCredential.cs,125

src/Couchbase.Lite.Shared/API/Sync/BasicAuthenticator.cs
```diff
@@ -50,15 +50,15 @@ public sealed class BasicAuthenticator : Authenticator
...
+        /// [Obsolete("This property is deprecated, please use SecureString PasswordSecureString instead.")]
         [NotNull]
         public string Password { get; }
 
         /// <summary>
         /// Gets the password that this object holds
         /// </summary>
         [NotNull]
-        internal SecureString PasswordSecureString { get; }
+        public SecureString PasswordSecureString { get; }
```
borrrden, line 61: Note that the use of SecureString is [no longer recommended by Microsoft](https://github.com/dotnet/platform-compat/blob/master/docs/DE0001.md).  Also this is a public API change, have you consulted @pasin on this?

src/LiteCore/src/LiteCore.Shared/Interop/Fleece.cs
borrrden: This ends up in a string in memory anyway so the question is: "Does it really provide more security than before?"

src/Couchbase.Lite.Shared/API/Sync/BasicAuthenticator.cs
pasin: Please use what is recommended by Microsoft.

borrrden (approved):
Discussed offline on the tradeoffs between this and other approaches

PR #1261, opened 2020-12-04
CBL-1452 Use c4db_getConfig2 in LiteCore binding tests

Sandychuang8:
CBL-1452 Use c4db_getConfig2 in LiteCore binding tests
- Add c4db get config2, delete name, open name, and copy name methods in c# binding
- Use the new methods shown above, and replaced the old ones in LiteCore binding tests

### DIFF
--- a/src/LiteCore/src/LiteCore.Shared/Interop/C4Database.cs
+++ b/src/LiteCore/src/LiteCore.Shared/Interop/C4Database.cs
@@ -110,96 +110,47 @@ namespace LiteCore.Interop
     }
 
     [ExcludeFromCodeCoverage]
-    internal sealed unsafe class DatabaseConfig2 : IDisposable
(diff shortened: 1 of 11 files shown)


src/LiteCore/src/LiteCore.Shared/Interop/C4Database.cs
```diff
@@ -108,4 +108,100 @@ public void Dispose()
             storageEngine = null;
         }
     }
+
+    [ExcludeFromCodeCoverage]
+    internal sealed unsafe class DatabaseConfig2 : IDisposable
```
borrrden, line 113: I'm unclear what the purpose of this class is at this point.  It doesn't seem to be doing anything substantial and I think it's going to lead to a lot of confusing corner cases later rather than keeping the abstraction thin like C4DatabaseConfig.

borrrden (approved):

PR #1262, opened 2020-12-07
CBL-1476 Add DNS support in proxy server address

Sandychuang8:
CBL-1476 Check if proxy server address is not IPAddress, then just past server host string into client connect async method.


borrrden (approved):
If you can simply pass a string DNS name in, I think you don't even need to parse the IP address in and you can just use the Host string unconditionally.  All of this needs testing though, either way.

PR #1267, opened 2021-01-26
CBL-1579 Replicator Heartbeat and test

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
```diff
@@ -195,6 +215,22 @@ public bool Reset
...
 
+        /// <summary>
+        /// Gets or sets the replicator heartbeat keep-alive interval.
+        /// </summary>
+        public TimeSpan Heartbeat
+        {
+            get => TimeSpan.FromSeconds(this.GetCast<long>(HeartbeatIntervalKey));
+            set { 
+                var sec = value.TotalSeconds;
+                if (sec > 0) {
```
borrrden, line 226: Be careful, `TotalSeconds` is a double type.  You should floor it first or 0.5, for example, will turn into 0 later and give unexpected results.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -58,6 +58,8 @@ public sealed unsafe class Replicator : IDisposable, IStoppable
             C4ReplicatorMode.Disabled, C4ReplicatorMode.Disabled, C4ReplicatorMode.OneShot, C4ReplicatorMode.Continuous
         };
 
+        private TimeSpan _defaultHeartbeatInterval = TimeSpan.FromSeconds(300);
```
borrrden, line 61: Why is this here?  Leftover?

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
```diff
@@ -37,23 +37,34 @@ internal sealed class ReplicatorOptionsDictionary : OptionsDictionary, IDisposab
...
         private const string PinnedCertKey = "pinnedCert";
         private const string OnlySelfSignedServerCert = "onlySelfSignedServer";
+
+        // WebSocket options:
         private const string ProtocolsOptionKey = "WS-Protocols";
-        private const string RemoteDBUniqueIDKey = "remoteDBUniqueID";
-        private const string ResetKey = "reset";
+        private const string HeartbeatIntervalKey = "heartbeat"; //Interval in secs to send a keepalive ping
+        
+        private TimeSpan _defaultHeartbeatInterval = TimeSpan.FromSeconds(300);
```
borrrden, line 66: Readability suggestion: `FromMinutes(5)`

Jenkinsfile
```diff
@@ -52,7 +52,7 @@ pipeline {
                         // }
                     }
                 }
-                stage("Linux Node") {
+                /*stage("Linux Node") {
```
borrrden, line 55: ??? What happened?

borrrden (changes requested):

Jenkinsfile
Sandychuang8: Linux pr validation tests are running flakily in 2.8.... and the fail test(s) are(is) not related to the pr itself, so I turned it off.

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
Sandychuang8: Change to value.Ticks/TimeSpan.TicksPerSecond. So when user assign TimeSpan.FromMilliseconds(800) to heartbeat interval, the invalid exception will throw.

borrrden: I know the error message is meant to be cross platform but it will be confusing that the user gets "Heartbeat Interval cannot be less or equal to 0."  Consider "Heartbeat Interval cannot be less or equal to 0 full seconds.

borrrden (approved):

PR #1268, opened 2021-02-16
CBL-1604 Retry Logic CBL-1645 Tests

Sandychuang8:


src/Couchbase.Lite.Shared/Support/Freezer.cs
```diff
@@ -58,6 +58,11 @@ public void SetValue<T>(ref T location, T newValue, [CallerMemberName]string cal
             location = newValue;
         }
 
+        public T GetValue<T>(ref T location)
+        {
+            return location;
```
borrrden, line 63: There is not really any reason to have this method, there is no harm in reading the variable directly

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -152,7 +152,14 @@ public IList<string> Channels
         public bool Continuous
         {
             get => _continuous;
-            set => _freezer.SetValue(ref _continuous, value);
+            set {
+                try {
+                    _freezer.SetValue(ref _continuous, value);
+                    Options.MaxRetries = _continuous ? Int32.MaxValue : 9;
+                } catch {
```
borrrden, line 159: What is this catching?

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
Sandychuang8: This is to take care the case when we alter the Continuous value in replicator configuration after replicator is created internally. 
```var config = CreateConfig(true, false, false);
            using (var repl = new Replicator(config)) {
                repl.Config.MaxRetries.Should().Be(9, "Because default Max Retries is 9 times for a Single Shot Replicator.");
                repl.Config.Continuous = true; // Internally
                repl.Config.MaxRetries.Should().Be(9, "Because default Max Retries is 9 times for a Single Shot Replicator.");
            }```

File: src/Couchbase.Lite.Shared/Support/Freezer.cs (hunk 1 above)
Comment (Sandychuang8) [reply src/Couchbase.Lite.Shared/Support/Freezer.cs:63]: Same reason as below. I want to retrieve the stored frozen Continuous value for internal test.

Comment (borrrden) [reply src/Couchbase.Lite.Shared/Support/Freezer.cs:63]: In the end all you are doing is returning `_continuous`.  The Freezer does not store anything, it's just a gatekeeper.

File: src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs (hunk 2 above)
Comment (borrrden) [reply src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs:159]: I don't understand.  What is the goal here?  Now the user will not know if they did something wrong.  Furthermore, if they set max retries before this will override it.

File: src/Couchbase.Lite.Shared/Support/Freezer.cs (hunk 1 above)
Comment (borrrden) [reply src/Couchbase.Lite.Shared/Support/Freezer.cs:63]: All this is doing is returning `_continuous` though, Freezer doesn't actually store anything, it's just returning the same thing that got passed in as the argument.

File: src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs (hunk 2 above)
Comment (borrrden) [reply src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs:159]: Now the user will not be alerted if they do something wrong.  What's wrong with not trying and catching?

File: src/Couchbase.Lite.Shared/Support/Freezer.cs (hunk 1 above)
Comment (Sandychuang8) [reply src/Couchbase.Lite.Shared/Support/Freezer.cs]: Removed.

File: src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs (hunk 2 above)
Comment (Sandychuang8) [reply src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs]: Logic removed.

File: src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs [hunk 3]
```diff
@@ -152,7 +152,10 @@ public IList<string> Channels
         public bool Continuous
         {
             get => _continuous;
-            set => _freezer.SetValue(ref _continuous, value);
+            set {
+                _freezer.SetValue(ref _continuous, value);
+                Options.MaxRetries = _continuous ? Int32.MaxValue : 9;
```
borrrden, line 157: I worry about this.  What if the user does this:

```
Config.MaxRetries = 5;
Config.Continuous = false;
```

Sandychuang8: Issue fixed.

Sandychuang8: But P2P recoverable tests are failing. Not sure if the change has caused the failures. Need to take a look.

Sandychuang8: After set the MaxRetries and MaxRetryWaitTime to the old default value 2 and 10 min, the P2P recoverable failure tests are passed.

borrrden (approved):

PR #1269, opened 2021-03-03
CBL-1681 Blob ToJSON, DB Save Blob and Get Blob CBL-1682 Document ToJSON and test, CBL-1687 Dictionary ToJSON and test

Sandychuang8:
Add New APIs (Blob.ToJSON, Database.SaveBlob, and Database.GetBlob)
Add tests

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/BlobTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/BlobTest.cs
@@ -174,7 +174,7 @@ namespace Test
         {
             var blob = ArrayTestBlob();
             Action badAction = (() => blob.ToJSON());
-            badAction.Should().Throw<InvalidDataException>(CouchbaseLiteErrorMessage.MissingDigestDueToBlobIsNotSavedToDB);
+            badAction.Should().Throw<InvalidOperationException>(CouchbaseLiteErrorMessage.MissingDigestDueToBlobIsNotSavedToDB);
 
             Db.SaveBlob(blob);
 
@@ -200,6 +200,11 @@ namespace Test
             var blobFromDict = Db.GetBlob(blobDict);
             blob.Equals(blobFromDict).Should().BeTrue();
 
+            //At this point Constants.ObjectTypeProperty key value pair is removed from the blobDict
+            Action badAction = (() => Db.GetBlob(blobDict));
+            badAction.Should().Throw<ArgumentException>(CouchbaseLiteErrorMessage.InvalidJSONDictionaryForBlob);
+
+            //Add back Constants.ObjectTypeProperty key value pair
             blobDict.Add(Constants.ObjectTypeProperty, "blob");
             blobDict.Remove(Blob.DigestKey);
             blobFromDict = Db.GetBlob(blobDict);
--- a/src/Couchbase.Lite.Tests.Shared/BlobTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/BlobTest.cs
@@ -32,6 +32,7 @@ using Couchbase.Lite.Util;
 using FluentAssertions;
 using LiteCore;
 using LiteCore.Interop;
+using Newtonsoft.Json;
 #if !WINDOWS_UWP
 using Xunit;
 using Xunit.Abstractions;
@@ -46,14 +47,14 @@ namespace Test
 #endif
     public sealed class BlobTest : TestCase
     {
-        #if !WINDOWS_UWP
+#if !WINDOWS_UWP
 
         public BlobTest(ITestOutputHelper output) : base(output)
         {
 
         }
 
-        #endif
+#endif
 
         [Fact]
         public void TestGetContent()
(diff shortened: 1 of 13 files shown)


src/Couchbase.Lite.Tests.Shared/BlobTest.cs
```diff
@@ -167,5 +168,85 @@ public unsafe void TestBlobStreamCopyTo()
...
+        {
+            var blob = ArrayTestBlob();
+            Db.SaveBlob(blob);
+
+            var blobDict = new Dictionary<string, object>() {
+                { Blob.ContentTypeKey, blob.ContentType },
+                { Blob.DigestKey, blob.Digest},
+                { Blob.LengthKey, blob.Length},
+                { Constants.ObjectTypeProperty, "blob" }
+            };
```
borrrden, line 198: This might be thinking a little too much but it might be good to do this via blob ToJSON instead so that if someone is looking at our tests for guidance on how to do this, they can see the supported method.  The flow that we are anticipating is:

1. Save the blob
2. Send the blob toJSON somewhere
3. Get the same JSON back later
4. Deserialize it and use it to look up the blob

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
```diff
@@ -2100,6 +2100,70 @@ public void TestRevisionIDExistingDoc()
             }
         }
 
+        [Fact]
+        public void TestTypesInDocumentToJSON()
+        {
+            var dic = PopulateDictData();
+            using (var md = new MutableDocument("doc1")) {
+                foreach (var item in dic) {
```
borrrden, line 2108: What is going on here?  Can't this just call `SetValue`?

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
```diff
@@ -2100,6 +2100,70 @@ public void TestRevisionIDExistingDoc()
...
+                }
+
+                Db.Save(md);
+            }
+
+            using (var doc = Db.GetDocument("doc1")) {
+                var json = doc.ToJSON();
+                var jdic = JsonConvert.DeserializeObject<DataInCBLDataType>(json);
+
+                VerifyValuesInJson(dic, jdic);
```
borrrden, line 2163: Isn't this testing that a blob is inside?  The ToJSON call should be returning a dictionary in place of a Blob but is that being tested?

src/Couchbase.Lite.Tests.Shared/BlobTest.cs
Sandychuang8: TestBlobToJSON starts from line 173 is very similar to what you describe in the steps.
Are you looking for replicator to be involved in step 2 and 3?

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
Sandychuang8: The test is actually to test Document ToJSON with all the CBL support types, and Blob is one of it.

Originally I have 2 branches, one blob ToJSON related (PR title reflects that) and the other one is Document and Dictionary ToJSON and it's branched from blob ToJSON. When I push the later branch, somehow it merged with the Blob PR...

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
Sandychuang8: I can also do SetValue with a dictionary, but since I am combining test types at once instead of each type individually, I want to keep using the specific set method for different type to distinguish from setValue with the whole dictionary....

src/Couchbase.Lite.Tests.Shared/BlobTest.cs
borrrden: No, just a start to end of the simulated flow for reference.  The replicator should not be involved here anyway, it's more of the transport between the library and third party functions.

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
borrrden: You can loop still, and just call `SetValue` right?  I don't see the need for these dozens of lines of code.

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
borrrden: From what I understood, the document JSON in ToJSON cannot have an actual blob inside of it, it needs to be all standard JSON types, which means the blob will be a dictionary.  That might be happening implicitly here but I can't tell.

Sandychuang8: Updated the VerifyValuesInJson to specifically run Blobs ToJSON and deserialize and compare both json values.

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
Sandychuang8: Done.

src/Couchbase.Lite.Tests.Shared/BlobTest.cs
Sandychuang8: Added a new blob test TestDbBlobToJson to show blob ToJSON behaves correctly during transport between the library and third party functions.

borrrden (approved):

PR #1270, opened 2021-03-06
CBL-1688 CBL-1684 CBL-1685 CBL-1683 CBL-1686 ToJSON API

Sandychuang8:
To JSON API
CBL-1688 Mutable Dictionary and tests
CBL-1684 Mutable Document and tests
CBL-1683 ArrayObject and test
CBL-1685 MutableArrayObject and tests
CBL-1686 Query Result and test


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -971,20 +971,17 @@ public void SaveBlob(Blob blob)
...
-                || (BlobDict.ContainsKey(Blob.ContentTypeKey) && BlobDict[Blob.ContentTypeKey].GetType() != typeof(string))
-                || (BlobDict.ContainsKey(Blob.LengthKey) && BlobDict[Blob.LengthKey].GetType() != typeof(int))
-                || BlobDict[Blob.DigestKey].GetType() != typeof(string)) {
+            if (!Blob.IsBlob(blobDict)) {
                 throw new ArgumentException(CouchbaseLiteErrorMessage.InvalidJSONDictionaryForBlob);
             }
 
             C4BlobKey expectedKey = new C4BlobKey();
-            var keyFromStr = Native.c4blob_keyFromString((string)BlobDict[Blob.DigestKey], &expectedKey);
+            var keyFromStr = Native.c4blob_keyFromString((string)blobDict[Blob.DigestKey], &expectedKey);
```
jayahariv, line 984: is this only for validating the size? since this is a lite core API to check for size and not mentioned in API spec. I am wondering is this necessary! 

@pasin  your thoughts?

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -282,6 +283,34 @@ internal Blob([NotNull]Database db, [NotNull]IDictionary<string, object> propert
...
+        /// length       | Number                 | No        | Binary length of the Blob in bytes.
+        /// digest       | String                 | Yes       | The cryptographic digest of the Blob’s content.
+        /// </param>
+        /// <returns>Return true if the given dictionary represents Blob, otherwise return false</returns>
+        public static bool IsBlob(Dictionary<string, object> blobDict)
+        {
+            if (!blobDict.ContainsKey(Constants.ObjectTypeProperty) || (string) blobDict[Constants.ObjectTypeProperty] != Constants.ObjectTypeBlob
+                || (blobDict.ContainsKey(Blob.ContentTypeKey) && blobDict[Blob.ContentTypeKey].GetType() != typeof(string))
+                || (blobDict.ContainsKey(Blob.LengthKey) && blobDict[Blob.LengthKey].GetType() != typeof(int))
+                || blobDict[Blob.DigestKey].GetType() != typeof(string)) {
```
jayahariv, line 305: in case of a Blob which is not saved to DB, this might fail right? since the digest is not present?

src/Couchbase.Lite.Shared/Document/InMemoryDictionary.cs
```diff
@@ -308,6 +310,12 @@ public IMutableDictionary SetValue(string key, object value)
             return this;
         }
 
+        public IMutableDictionary SetJSON([NotNull] string json)
+        {
+            var blobDict = JsonConvert.DeserializeObject<Dictionary<string, object>>(json);
```
jayahariv, line 315: exception handling?

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: In line 976, I have the logic checking that. Since we still allow blob to be created without a digest, so I am not sure if I should add that logic in IsBlob logic.....

jayahariv (commented):
Looks good!

src/Couchbase.Lite.Shared/API/Database/Database.cs
pasin: What is the purpose of getting key from string here?

src/Couchbase.Lite.Shared/API/Document/MutableDocument.cs
```diff
@@ -347,6 +361,24 @@ public IMutableDictionary SetDictionary(string key, DictionaryObject value)
...
+        public IMutableDictionary SetJSON([NotNull] string json)
+        {
+            JObject jobj = null;
+            try {
+                jobj = JObject.Parse(json);
+            } catch {
+                throw new CouchbaseLiteException(C4ErrorCode.InvalidParameter, CouchbaseLiteErrorMessage.InvalidJSON);
+            }
+
+            if (jobj.GetType() == typeof(JArray)) {
```
pasin, line 374: Should it check if it's not dictionary instead? The other types such as string and numbers are also valid JSON.

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -167,6 +169,24 @@ internal static object ToNetObject(object value)
             }
         }
 
+        internal static object ToJsonObject(object value)
+        {
+            switch (value) {
+                case null:
+                    return null;
+                case InMemoryDictionary inMem:
+                    return inMem.ToDictionary();
```
pasin, line 178: How do Blob objects get handled?

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -167,6 +169,24 @@ internal static object ToNetObject(object value)
...
+                case IDictionaryObject roDic:
+                    return roDic.ToDictionary();
```
pasin, line 180: How do Blob objects get handled?

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -167,6 +169,24 @@ internal static object ToNetObject(object value)
...
+                case IArray roarr:
+                    return roarr.ToList();
```
pasin, line 182: How do Blob objects get handled?

src/Couchbase.Lite.Shared/API/Document/Blob.cs
pasin: From the public API point of view, digest is required.

src/Couchbase.Lite.Shared/API/Document/Document.cs
```diff
@@ -403,6 +404,10 @@ public void Dispose()
         /// <inheritdoc />
         public string ToJSON()
         {
+            if(IsMutable) {
+                return JsonConvert.SerializeObject(_dict);
```
pasin, line 408: How does this handle Blob objects?

pasin (commented):
When converting Document/Dictionary/Array (Mutable or Immutable), should we do it the same way that we do Fleece Encoding so that Blob can be handled?

src/Couchbase.Lite.Shared/API/Document/Blob.cs
Sandychuang8: The GetBlob API (line 974) will return null if the digest info doesn't existed as expected in spec. If I add digest null checking into IsBlob, which returns bool, and GetBlob throws exception if IsBlob returns false. No digest will returns false in isBlob, we then will not be able to get null result from GetBlob when there is no digest existed.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: We have a test case in GetBlob to test with digest with a random string as value. We need to use the random string to get the C4Blob key and then use the key we get the check if c4blob_getSize returns a valid value, otherwise null if -1 is returned.

src/Couchbase.Lite.Shared/Document/InMemoryDictionary.cs
Sandychuang8: JsonConvert.DeserializeObject will throw json string is unable to deserialize into a dictionary type.

src/Couchbase.Lite.Shared/API/Database/Database.cs
pasin: See, I didn't see get size for some reason. Thanks.

src/Couchbase.Lite.Shared/API/Document/Document.cs
Sandychuang8: I have made modification to the Blob class to make it serializable to Dictionary<string, object> json string represents Blob and Dictionary<string, object> represents Blob deserializable.

src/Couchbase.Lite.Shared/API/Document/MutableDocument.cs
Sandychuang8: Line 378 below will deserialize json string into Dictionary<string, object> object, which will make the validation there.
JObject.Parse(json); checks if string passed in is json format, and jobj  is the json object if json successfully parsed. Then we will need to make sure jobj is not an array type before we deserialize the json string into dictionary object.

src/Couchbase.Lite.Shared/Document/DataOps.cs
Sandychuang8: ToJsonObject(object value) will convert Blob object into json representation dictionary object.

src/Couchbase.Lite.Shared/Document/DataOps.cs
Sandychuang8: When Blob is passed in as value parameter, it will go to case Blob and handle from there.

src/Couchbase.Lite.Shared/Document/DataOps.cs
Sandychuang8: Same thing, when a Blob is passed in as value parameter, it will go to case Blob and handle from there.

src/Couchbase.Lite.Shared/API/Document/MutableDictionaryObject.cs
```diff
@@ -219,6 +234,24 @@ public IMutableDictionary SetDictionary(string key, DictionaryObject value)
...
+                jobj = JObject.Parse(json);
+            } catch {
+                throw new CouchbaseLiteException(C4ErrorCode.InvalidParameter, CouchbaseLiteErrorMessage.InvalidJSON);
+            }
+
+            if (jobj.GetType() == typeof(JArray)) {
+                throw new CouchbaseLiteException(C4ErrorCode.InvalidParameter, CouchbaseLiteErrorMessage.InvalidJSON);
+            }
+
+            var blobDict = JsonConvert.DeserializeObject<Dictionary<string, object>>(json);
```
borrrden, line 251: This can throw if I recall correctly so I think it should be wrapped in try / catch that will throw the same exception as above.

src/Couchbase.Lite.Shared/API/Document/MutableDocument.cs
borrrden: What will happen if it is just a string or number like Pasin says though?

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -110,6 +110,8 @@ internal static object ToCouchbaseObject(object value)
                     return null;
                 case DateTimeOffset dto:
                     return dto.ToString("o");
+                case DateTime dt: //json serialized datetimeoffset to date time so convert it dto
+                    return new DateTimeOffset(dt.ToUniversalTime());
```
borrrden, line 114: There is a setting to control this, you should make use of it like other classes in Couchbase Lite

src/Couchbase.Lite.Shared/Document/InMemoryDictionary.cs
borrrden: That's the point, it should be wrapped in a CBL exception, with this one as the inner one.

borrrden (changes requested):

src/Couchbase.Lite.Shared/Document/DataOps.cs
pasin: I guessed, I couldn't see how this work recursively from the code in the PR. Please make sure if the document / dictionary / array with Blob inside the nested dictionary also works.

For example, the document with the following structure,
{
  "blob: blob1,
   "subDict": { "blob": blob2 }, 
   "subArray": [ blob3 ], 
}

Note: blob1, blob2, and blob3 are the actual Blob objects.

src/Couchbase.Lite.Shared/Document/InMemoryDictionary.cs
Sandychuang8: done

src/Couchbase.Lite.Shared/API/Document/MutableDocument.cs
Sandychuang8: It throws invalid argument. I just added couple new tests TestMutableDictSetJsonWithInvalidParam & TestMutableDocumentSetJsonWithInvalidParam to prove.

src/Couchbase.Lite.Shared/Document/DataOps.cs
Sandychuang8: I will add more test cases to validate.

src/Couchbase.Lite.Shared/API/Document/MutableDictionaryObject.cs
Sandychuang8: done.

src/Couchbase.Lite.Shared/API/Document/MutableDocument.cs
borrrden: 👍

src/Couchbase.Lite.Shared/Document/DataOps.cs
Sandychuang8: I added TestBlobJsonStringInLayersOfMutableDict to validate blobs are retriable and values are correct from subarray and sub dictionary in a mutable array.

src/Couchbase.Lite.Tests.Shared/ArrayTest.cs
```diff
@@ -1367,6 +1370,133 @@ public void TestTypes()
...
+                        break;
+                    case DateTimeOffset dto:
+                        ma.GetDate(index).Should().Be(dto);
+                        break;
+                    case byte b:
+                    case ushort us:
+                    case uint ui:
+                    case ulong ul:
+                        Equals(Convert.ToUInt64(ma.GetValue(index)), Convert.ToUInt64(array[index])).Should().BeTrue();
+                        break;
```
borrrden, line 1433: "Equals should be true" is not a very fluent use of the assertions.

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
```diff
@@ -2126,9 +2143,55 @@ public void TestMutableDocWithJsonString()
             var dic = PopulateDictData();
             var dicJson = JsonConvert.SerializeObject(dic);
             using (var md = new MutableDocument("doc1", dicJson)) {
-                var mdJson = md.ToJSON();
-                var mdJsonDict = JsonConvert.DeserializeObject<Dictionary<string, object>>(mdJson);
-                VerifyValuesInJson(dic, mdJsonDict);
+                foreach (var kvPair in dic) {
```
borrrden, line 2128: Is this just copied and pasted from above?

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -120,12 +120,13 @@ internal static object ToCouchbaseObject(object value)
                     var jobjDict = jobj.ToObject<IDictionary<string, object>>();
                     if (jobjDict.ContainsKey(Constants.ObjectTypeProperty) && 
                         jobjDict[Constants.ObjectTypeProperty].Equals(Constants.ObjectTypeBlob) &&
-                        jobjDict.ContainsKey("Content") && jobjDict["Content"] != null &&
+                        jobjDict.ContainsKey(Blob.ContentTypeKey) && jobjDict[Blob.ContentTypeKey] != null &&
                         jobjDict.ContainsKey("Content") && jobjDict["Content"] != null) {
```
borrrden, line 123: Is `Blob.ContentTypeKey` different from "Content" ?

borrrden (commented):
This PR has gotten absolutely gargantuan and is a good example of why not to take on 5 different tickets at once :(

src/Couchbase.Lite.Shared/Document/DataOps.cs
borrrden: I still notice that you are handling `DateTime` here when we didn't before.  Everything should be using `DateTimeOffset`

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
Sandychuang8: yeah... it's very very similar, but one is mutable dictionary and the other one is mutable document...

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -154,6 +154,15 @@ public void Set(int index, object val)
 
         #endregion
 
+        #region Internal Methods
+
+        internal string ToJSON()
```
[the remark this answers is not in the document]
pasin: I think it would be good to step back and synchronize how to implement ToJSON() with iOS and Android platform.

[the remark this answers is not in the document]
Sandychuang8: This internal ToJSON() will be called by none mutable array object, Line 224 under ArrayObject class will check if the the array object is mutable or not.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -280,6 +282,47 @@ internal Blob([NotNull]Database db, [NotNull]IDictionary<string, object> propert
...
+        /// ---------------------------------------------------------------------------------------------------
+        /// @type        | constant string “blob” | Yes       | Indicate Blob data type.
+        /// content_type | String                 | No        | Content type ex. text/plain.
+        /// length       | Number                 | No        | Binary length of the Blob in bytes.
+        /// digest       | String                 | Yes       | The cryptographic digest of the Blob’s content.
+        /// </param>
+        /// <returns>Return true if the given dictionary represents Blob, otherwise return false</returns>
+        public static bool IsBlob(IDictionary<string, object> blobDict)
+        {
+            if (!blobDict.ContainsKey(Constants.ObjectTypeProperty) || (string) blobDict[Constants.ObjectTypeProperty] != Constants.ObjectTypeBlob
```
borrrden, line 316: Will throw an invalid cast exception if it is not a string.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
```diff
@@ -291,13 +334,18 @@ internal void FLEncode(FLEncoder* enc)
                 // This blob is attached to a document, so save the full metadata
                 var document = GCHandle.FromIntPtr((IntPtr) extra).Target as MutableDocument;
                 var database = document.Database;
-                try {
-                    Install(database);
-                } catch (Exception) {
-                    WriteLog.To.Database.W(Tag, "Error installing blob to database, throwing...");
-                    throw;
+                if (Digest != null && _db == null) {
+                    _db = database;
```
borrrden, line 338: I have a feeling this logic would be better off inside of `Install` but it works here too as long as there are not other places which also need it.

src/Couchbase.Lite.Shared/API/Document/DictionaryObject.cs
```diff
@@ -55,7 +55,7 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             }
 
             while (reader.TokenType != JsonToken.EndObject && reader.Read()) {
-                var key = reader.Value as string;
+                var key = reader.Path as string;
```
borrrden, line 58: What is this change for?  What does `Path` return?  I worry it is the full path from the root of the JSON and not just the value of the current key.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -943,24 +943,13 @@ public bool Save([@NotNull]MutableDocument document, [@NotNull]Func<MutableDocum
         /// <param name="blob">The blob object will be saved into Database.</param>
         public void SaveBlob(Blob blob)
```
borrrden, line 944: I really want to see warnings all over this to not use it for general cases ( @pasin )

src/Couchbase.Lite.Tests.Shared/DocumentTest.cs
borrrden: Both of them implement `IMutableDictionary` so I think you can refactor into another method at a later point.

borrrden (changes requested):
The warnings in doc comments, unhandled exception are what I marked this red for.  The `reader.Value` -> `reader.Path` I want confirmation of intention and the others are optional.

src/Couchbase.Lite.Shared/API/Document/DictionaryObject.cs
Sandychuang8: I reverted the change. The reason why I update it to Path was because I got a null reference error when I played with serialize/deserialize the DictionaryObject from/to json, and noticed the Value was null while the Path was returning the correct value. This ReadJson requires a 2nd look when we work on Data Model feature.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
Sandychuang8: Moved.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
Sandychuang8: Add an additional check if blobDict[Constants.ObjectTypeProperty] is string before cast it.

src/Couchbase.Lite.Shared/API/Document/ArrayObject.cs
```diff
@@ -220,6 +219,20 @@ IEnumerator IEnumerable.GetEnumerator()
...
         #endregion
 
+        #region IJSON
+
+        /// <inheritdoc />
+        public string ToJSON()
+        {
+            if (_array.IsMutable) {
+                throw new NotSupportedException();
+            } else {
```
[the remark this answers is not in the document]
Sandychuang8: Removed.

src/Couchbase.Lite.Shared/API/Document/Blob.cs
borrrden: This will do, `as` could have compared them in one line though

borrrden (approved):
Still want to see warnings on the manual blob API so that developers not making a plugin don't get any clever ideas....

src/Couchbase.Lite.Shared/API/Database/Database.cs
pasin: Let's do in the the API spec so others could see it too. At the very least, the API doc should say something.

PR #1279, opened 2021-04-29
CBL-1565 Ensure c4log_enableFatalExceptionBacktrace is called

Sandychuang8:
CBL-1565 Ensure c4log_enableFatalExceptionBacktrace is called


borrrden (approved):
Note:  Since this is a draft I won't be getting pinged about it.

PR #1282, opened 2021-05-04
CBL-1909 Save Cookie and Get Cookies will now only log errors.

Sandychuang8:
Save Cookie and Get Cookies will now only log error instead of throw exceptions.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -1060,12 +1060,13 @@ namespace Couchbase.Lite
                     addr.port = (ushort) uri.Port;
                     addr.path = path.AsFLSlice();
 
-                    cookies = (string) LiteCoreBridge.Check(err =>
-                    {
-                        return Native.c4db_getCookies(_c4db, addr, err);
-                    });
+                    C4Error err = new C4Error();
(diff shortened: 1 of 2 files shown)


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1060,12 +1060,13 @@ internal string GetCookies([@NotNull] Uri uri)
...
-                        return Native.c4db_getCookies(_c4db, addr, err);
-                    });
+                    C4Error err = new C4Error();
+                    cookies = Native.c4db_getCookies(_c4db, addr, &err);
+                    if (err.code > 0) {
+                        WriteLog.To.Sync.W(Tag, $"{err.domain}/{err.code} Failed getting Cookie from address {addr}.");
+                    }
 
-                    if (String.IsNullOrEmpty(cookies)) {
+                    if (String.IsNullOrEmpty(cookies) || err.code > 0) {
```
borrrden, line 1069: I don't think that this needs this extra condition here.  Isn't it the case that if the error code is non zero cookies will always be null?  I think the right check here is `if (String.IsNullOrEmpty(cookies) && err.code == 0)` because that means that the call succeeded and no cookies were found.

borrrden (changes requested):

borrrden (approved):

PR #1283, opened 2021-05-05
CBL-1905 Use the setProgressLevel API

Sandychuang8:
Passing in progress level via configuration is deprecated; use the setProgressLevel API

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -197,6 +197,10 @@ namespace Couchbase.Lite.Sync
         {
             CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(handler), handler);
             var cbHandler = new CouchbaseEventHandler<DocumentReplicationEventArgs>(handler, scheduler);
+            if (_repl!= null && _documentEndedUpdateEventsCount <= 0) {
+                SetProgressLevel(C4ReplicatorProgressLevel.PerDocument);
+            }
+            
             _documentEndedUpdate.Add(cbHandler);
             _documentEndedUpdateEventsCount++;
             return new ListenerToken(cbHandler, "repl");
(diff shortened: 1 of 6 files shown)


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -746,6 +752,14 @@ private C4Error SetupC4Replicator()
                 else
             #endif
                     _repl = Native.c4repl_new(Config.Database.c4db, addr, dbNameStr, _nativeParams.C4Params, &localErr);
+
+                if (_documentEndedUpdateEventsCount > 0) {
+                    C4Error err = new C4Error();
+                    if (!Native.c4repl_setProgressLevel(_repl, C4ReplicatorProgressLevel.PerDocument, &err) || err.code > 0) {
+                        WriteLog.To.Sync.W(Tag, "Failed set progress level to PerDocument", err);
+                    }
+                }
```
borrrden, line 761: I don't remember the nuance here, but why is this logic here instead of when adding the change listener.  What if you add a change listener after starting?

Sandychuang8: c4 replicator would be null if I call setProgressLevel in the AddChangeListener. That is why I move the call down here after the c4 replicator is created.

borrrden: That doesn't account for the situation where a change listener is added after start though.  I think it is needed there too with a null check.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -195,9 +196,13 @@ public ListenerToken AddDocumentReplicationListener([NotNull]EventHandler<Docume
             [NotNull]EventHandler<DocumentReplicationEventArgs> handler)
         {
             CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(handler), handler);
-            Config.Options.ProgressLevel = ReplicatorProgressLevel.PerDocument;
             var cbHandler = new CouchbaseEventHandler<DocumentReplicationEventArgs>(handler, scheduler);
+            if (_repl!= null && _documentEndedUpdateEventsCount <= 0) {
```
borrrden, line 200: This is starting to break thread safety.  The `Event` class here already reports how many events are present when adding and removing.  I think it would be safer to add a method to get the current count from inside the event rather than tracking it separately like this.

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
Sandychuang8: The count is not accessible in Event. I think I did try that first.

borrrden (approved):

PR #1285, opened 2021-05-13
CBL-1865 Enhanced Configuration API in struct CBL-1912 Update Retry and Heartbeat logic

Sandychuang8:
Enhanced Configuration (immutable configuration)
- Database config
- Replicator config
- URLEndpoint config

CBL-1912 Update Retry and heartbeat logic

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/DatabaseConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/DatabaseConfiguration.cs
@@ -1,7 +1,7 @@
 ﻿// 
 //  DatabaseConfiguration.cs
 //
-//  Copyright (c) 2017 Couchbase, Inc All rights reserved.
+//  Copyright (c) 2021 Couchbase, Inc All rights reserved.
 // 
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
@@ -17,9 +17,6 @@
 // 
 
 using Couchbase.Lite.DI;
-using Couchbase.Lite.Internal.Logging;
-using Couchbase.Lite.Support;
-using Couchbase.Lite.Util;
 
 using JetBrains.Annotations;
 
@@ -28,7 +25,7 @@ namespace Couchbase.Lite
     /// <summary>
     /// A struct containing configuration for creating or opening database data
     /// </summary>
-    public sealed class DatabaseConfiguration
+    public readonly struct DatabaseConfiguration
     {
         #region Constants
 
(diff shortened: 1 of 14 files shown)


src/Couchbase.Lite.Shared/API/Database/DatabaseConfiguration.cs
```diff
@@ -69,29 +69,30 @@ public string Directory
...
         /// <summary>
         /// Default constructor
         /// </summary>
-        public DatabaseConfiguration()
+        /// <param name="directory">
+        /// Default directory is <see cref="Service.GetRequiredInstance<IDefaultDirectoryResolver>().DefaultDirectory()" /> if directory set to null.
+        /// </param>
+        public DatabaseConfiguration(
+#if COUCHBASE_ENTERPRISE
+            EncryptionKey encryptionKey = null,
```
borrrden, line 84: Let's have this as the last argument instead of the first so that if the user says `new DatabaseConfiguration(directory)` it will work in both CE and EE.

src/Couchbase.Lite.Shared/API/Database/DatabaseConfiguration.cs
```diff
@@ -28,7 +25,7 @@ namespace Couchbase.Lite
     /// <summary>
     /// A struct containing configuration for creating or opening database data
     /// </summary>
-    public sealed class DatabaseConfiguration
+    public readonly struct DatabaseConfiguration
```
borrrden, line 28: I've never actually gotten to use readonly struct before!  This will be nice!

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -119,7 +119,7 @@ static Replicator()
         /// <param name="config">The configuration to use to create the replicator</param>
         public Replicator([NotNull]ReplicatorConfiguration config)
         {
-            Config = CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(config), config);
+            Config = config;// CBDebug.MustNotBeNull(WriteLog.To.Sync, Tag, nameof(config), config);
```
borrrden, line 122: Don't leave in the comment

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -76,184 +76,164 @@ public enum DocumentFlags
...
+        internal const long DefaultMaxRetryInterval = 300;
+        internal const int MaxRetriesContinuous = int.MaxValue;
+        internal const int MaxRetriesOneShot = 9;
+
         #endregion
 
         #region Variables
 
-        [NotNull]private readonly Freezer _freezer = new Freezer();
+        private readonly Freezer _freezer;
```
borrrden, line 94: This is no longer needed.

borrrden (changes requested):

Sandychuang8:
Will not proceeding creating configuration in struct. Enhanced configuration with init only property will be worked on when we officially switch to C# 9.

PR #1289, opened 2021-05-26
CBL-1886 Fix couple null checking caused resolver returns null tests failures

Sandychuang8:


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1509,7 +1509,6 @@ private bool SaveResolvedDocument([@CanBeNull]Document resolvedDoc, [@NotNull]Do
 
             // mergedBody:
             FLSliceResult mergedBody = (FLSliceResult) FLSlice.Null;
-            if (resolvedDoc != null)
             if (!ReferenceEquals(resolvedDoc, remoteDoc)) {
```
Sandychuang8, line 1512: Not sure why `if (resolvedDoc != null)` is added here.. Remove.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1522,7 +1521,7 @@ private bool SaveResolvedDocument([@CanBeNull]Document resolvedDoc, [@NotNull]Do
             }
 
             // mergedFlags:
-            C4RevisionFlags mergedFlags = resolvedDoc.c4Doc != null ? resolvedDoc.c4Doc.RawDoc->selectedRev.flags : 0;
+            C4RevisionFlags mergedFlags = resolvedDoc?.c4Doc != null ? resolvedDoc.c4Doc.RawDoc->selectedRev.flags : 0;
```
Sandychuang8, line 1524: Since resolvedDoc can be null so make is nullable in conditional expression.

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: Looks like a merge error to me

borrrden (approved):

PR #1290, opened 2021-06-04
CBL-1793 Change to QueryBuilder API

Sandychuang8:
1. FTS Match Functions
- Deprecate FullTextExpression class and it's Index Match full text expression,  IFullTextExpression.
- Add Index name Match query expression function in FullTextFunction class.
- Update tests
2. isValued and isNotValued operator
- Deprecate IsNullOrMissing and NotNullOrMissing from QueryExpression and IExpression.
- Add IsValued and IsNotValued in QueryExpression and IExpression.
- Update tests

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -405,42 +405,6 @@ namespace Test
         [Fact]
         public void TestReplicatorOneMaxAttempts() => ReplicatorMaxAttempts(1);
 
-        //[Fact]
-        public void TestReplicatorMaxAttemptsNoneContinuous()
-        {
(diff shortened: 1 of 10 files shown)


src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -405,6 +405,42 @@ public void TestReplicatorMaxAttemptsGetSet()
         [Fact]
         public void TestReplicatorOneMaxAttempts() => ReplicatorMaxAttempts(1);
 
+        //[Fact]
+        public void TestReplicatorMaxAttemptsNoneContinuous()
```
borrrden, line 409: Why is this getting added as part of this PR?

borrrden (approved):

PR #1292, opened 2021-06-08
CBL-1790 Create Query with N1QL string

Sandychuang8:
* Use new c4query_new2 instead of c4query_new.
* Add CreateQuery method in Database class

### DIFF
--- a/src/Couchbase.Lite.Shared/Query/N1QLQuery.cs
+++ b/src/Couchbase.Lite.Shared/Query/N1QLQuery.cs
@@ -1,0 +1,121 @@
+﻿using Couchbase.Lite.Query;
+using Couchbase.Lite.Support;
+using JetBrains.Annotations;
+using LiteCore.Interop;
+using System;
+using System.Collections.Generic;
+
+namespace Couchbase.Lite.Internal.Query
+{
+    internal class N1QLQuery : XQuery
+    {
+        private const string Tag = nameof(N1QLQuery);
+        private string _n1qlQueryExpression = "";
+
+        [NotNull]
+        internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
+
+        public N1QLQuery(string n1qlQueryExpression, Database database) : base()
+        {
+            Database = database;
+            _n1qlQueryExpression = n1qlQueryExpression;
+        }
+
+        public override unsafe IResultSet Execute()
+        {
+            if (Database == null) {
+                throw new InvalidOperationException(CouchbaseLiteErrorMessage.InvalidQueryDBNull);
+            }
+
+            var options = C4QueryOptions.Default;
+            var paramN1ql = _n1qlQueryExpression.FLEncode();
+
+            var e = (C4QueryEnumerator*)ThreadSafety.DoLockedBridge(err =>
+            {
+                if (_disposalWatchdog.IsDisposed) {
+                    return null;
+                }
+
+                if (_c4Query == null) {
+                    Check();
+                }
+
+                var localOpts = options;
+                return NativeRaw.c4query_run(_c4Query, &localOpts, FLSlice.Null, err);
+            });
+
+            paramN1ql.Dispose();
+
+            if (e == null) {
+                return new NullResultSet();
+            }
+
+            var retVal = new QueryResultSet(this, ThreadSafety, e, _columnNames);
+            _history.Add(retVal);
+            return retVal;
+        }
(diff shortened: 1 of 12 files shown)


src/Couchbase.Lite.Shared/Query/XQuery.cs
```diff
@@ -48,13 +48,13 @@ internal class XQuery : IQuery, IStoppable
         #region Variables
 
         [NotNull]private readonly Event<QueryChangedEventArgs> _changed = new Event<QueryChangedEventArgs>();
-        [NotNull] private readonly DisposalWatchdog _disposalWatchdog = new DisposalWatchdog(nameof(IQuery));
+        [NotNull] internal readonly DisposalWatchdog _disposalWatchdog = new DisposalWatchdog(nameof(IQuery));
```
borrrden, line 51: Why not protected?

src/Couchbase.Lite.Shared/Query/N1QLQuery.cs
```diff
@@ -0,0 +1,121 @@
+﻿using Couchbase.Lite.Query;
+using Couchbase.Lite.Support;
+using JetBrains.Annotations;
+using LiteCore.Interop;
+using System;
+using System.Collections.Generic;
+
+namespace Couchbase.Lite.Internal.Query
+{
+    internal class N1QLQuery : XQuery
```
borrrden, line 10: Not really wild about this inheritance scheme.  A N1QL query is not really a specialization of XQuery, but a sibling to it.  I think this could be refactored better, but I will just leave this as a comment.

Sandychuang8: Created a QueryBase class to share between XQuery and NQuery.

src/Couchbase.Lite.Shared/Query/NQuery.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        #region Override Methods
+        public override unsafe IResultSet Execute()
+        {
+            if (Database == null) {
+                throw new InvalidOperationException(CouchbaseLiteErrorMessage.InvalidQueryDBNull);
+            }
+
+            var options = C4QueryOptions.Default;
+            var paramN1ql = _n1qlQueryExpression.FLEncode();
```
borrrden, line 56: What is this unused variable for?

src/Couchbase.Lite.Shared/Query/NQuery.cs
```diff
@@ -0,0 +1,155 @@
...
+                    _columnNames = CreateColumnNames(query);
+                }
+
+                Native.c4query_release(_c4Query);
+                _c4Query = query;
+                return true;
+            });
+        }
+
+        private unsafe Dictionary<string, int> CreateColumnNames(C4Query* query)
```
borrrden, line 136: Is there anything different here vs XQuery?

src/Couchbase.Lite.Shared/Query/NQuery.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            var retVal = new QueryResultSet(this, ThreadSafety, e, _columnNames);
+            _history.Add(retVal);
+            return retVal;
+        }
+
+        public override string Explain()
+        {
+            return _n1qlQueryExpression;
```
borrrden, line 85: I don't think this is correct, why is it not the same as XQuery?

src/Couchbase.Lite.Shared/Query/NQuery.cs
Sandychuang8: N1QL query doesn't have FromImpl
`
var fromImpl = FromImpl;
Debug.Assert(fromImpl != null, "CreateColumnNames reached without a FROM clause received");
`

src/Couchbase.Lite.Shared/Query/NQuery.cs
Sandychuang8: removed

src/Couchbase.Lite.Shared/Query/NQuery.cs
Sandychuang8: revert to use c4query_explain(_c4Query) for query explain

borrrden (approved):
It's not good to have more than one PR for the same JIRA ticket number because the first one that gets merged will close the ticket

PR #1293, opened 2021-06-09
CBL-1790 Create Query Index with N1QL string array

Sandychuang8:
- Use new c4db_createIndex2 instead of c4db_createIndex
- Add CreateIndex method takes IndexConfiguration as parameter in Database class
- Add IndexConfiguration class with value index and full-text index creation

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Query/IIndexConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Query/IIndexConfiguration.cs
@@ -1,0 +1,57 @@
+﻿// 
+// IIndexConfiguration.cs
+// 
+// Copyright (c) 2021 Couchbase, Inc All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


src/Couchbase.Lite.Shared/API/Query/IIndexConfiguration.cs
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// 
+
+using JetBrains.Annotations;
+
+namespace Couchbase.Lite.Query
+{
+    public interface IIndexConfiguration
```
borrrden, line 23: I was left with the impression that we wouldn't need to have this sort of interface / factory system.  Is there a benefit to having this rather than having it rather than just defining classes directly?

Sandychuang8: I don't want to change the existing IIndex, IndexBuilder structure and in order for IndexBuilder and IndexConfiguration to reuse IndexExplainer, IIndexConfiguration, IndexConfiguration need to be in similar structure of IIndex, IndexBuilder.

So IndexConfiguration can do something similar like `IndexConfiguration.FullTextIndex("es_detail").IgnoreAccents(true).SetLanguage("es");`

borrrden: @pasin Is this API compliant with the spec?

pasin: The IndexConfiguration API is in the form of the configuration class. It should be like any other configuration class. We don't use the factory pattern here. Please check the API spec.

Sandychuang8:
Need to redo Create Query Index with N1QL string array

PR #1294, opened 2021-06-10
CBL-1790 Create Query Index with N1QL string array

Sandychuang8:
- Use new c4db_createIndex2 instead of c4db_createIndex
- Add CreateIndex method takes IndexConfiguration as parameter in Database class
- Add ValueIndexConfiguration class and FullTextIndexConfiguration
- Update LiteCore to Commit: 63d918b967dbf9170c970381a3eacd9f4eef2a33 [63d918b] (Add the capability of createIndex to take index expression in N1QL)

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Query/FullTextIndexConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Query/FullTextIndexConfiguration.cs
@@ -1,0 +1,87 @@
+﻿// 
+// FullTextIndexConfiguration.cs
+// 
+// Copyright (c) 2021 Couchbase, Inc All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// 
+
+using Couchbase.Lite.Internal.Query;
+using LiteCore.Interop;
+using System.Globalization;
+
+namespace Couchbase.Lite.Query
+{
+    /// <summary>
+    /// An class for an index based on full text searching
+    /// </summary>
+    public sealed class FullTextIndexConfiguration : IndexConfiguration
+    {
+        #region Variables
+
+        private bool _ignoreAccents;
+        private string _locale = CultureInfo.CurrentCulture.TwoLetterISOLanguageName;
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Gets whether or not to ignore accents when performing 
+        /// the full text search
+        /// </summary>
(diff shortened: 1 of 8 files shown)


src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Couchbase.Lite.Internal.Query
+{
+    public abstract class IndexConfiguration
+    {
+        #region Properties
+
+        protected string[] Expression { get; }
+
+        internal C4QueryLanguage QueryLanguage { get; }
```
borrrden, line 30: I don't think these are meant to be internal?

Sandychuang8: hmm? Do we need to let the developers know these c4 query language types: C4QueryLanguage.N1QLQuery or C4QueryLanguage.JSONQuery?

borrrden: Oh nevermind I thought these were the public properties that Pasin mentioned (I saw "expression" that looked familiar) but they aren't right?

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        internal C4IndexType IndexType { get; }
+
+        internal abstract C4IndexOptions Options { get; }
+
+        #endregion
+
+        #region Constructor
+
+        internal IndexConfiguration(C4IndexType indexType, params string[] items)
```
borrrden, line 40: Best practice for abstract classes is protected constructors

src/Couchbase.Lite.Shared/API/Query/FullTextIndexConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Gets whether or not to ignore accents when performing 
+        /// the full text search
+        /// </summary>
+        public bool IgnoreAccents => _ignoreAccents;
```
borrrden, line 43: Why not auto property?

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
borrrden: Just confirmed with @pasin That "Expressions" should be public according to the spec.

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
Sandychuang8: Sure, but in this case the C4IndexType is internal and I will get inconsistent accessibility compiler error..

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
borrrden: It is "Expressions" (with an S) not "Expression"

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
borrrden: Ok, that's a good enough reason then.

borrrden (approved):

PR #1295, opened 2021-06-12
CBL-2008 Update LiteCore Dev branch

Sandychuang8:
- Remove db compact, starthousekeeping APIs, and update push/pull filter callback with additional collectionName parameter.
- C4Document struct update.
- Update Create Query with n1ql From clause in collection.
- Use getIndexesInfo and deprecate getIndexes


borrrden (approved):
Need to file tickets about the tests that are now disabled

PR #1297, opened 2021-06-22
CBL-1923 Construct From data source JSON AS db name

Sandychuang8:
The existing API shouldn't change, so revert test and update the From clause ToJSON.


borrrden (approved):
I assume there is still a test which checks that the `As` function still results in the proper name in the query results

Sandychuang8:
Existing As function tests are still passing.

PR #1300, opened 2021-06-30
Check Replicator Activity level

Sandychuang8:


src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -246,12 +246,17 @@ public void Start(bool reset)
...
-                    Config.Database.AddActiveStoppable(this);
                     status = Native.c4repl_getStatus(_repl);
+                    if (status.level == C4ReplicatorActivityLevel.Stopped
+                    || status.level == C4ReplicatorActivityLevel.Stopping
+                    || status.level == C4ReplicatorActivityLevel.Offline) {
+                        ServerCertificate = null;
+                        WriteLog.To.Sync.I(Tag, $"{this}: Starting");
+                        Native.c4repl_start(_repl, Config.Options.Reset || reset);
+                        Config.Options.Reset = false;
+                        Config.Database.AddActiveStoppable(this);
```
Sandychuang8, line 257: Actually AddActiveStoppable is no-op if the replicator is already in ActiveStoppable..

borrrden (commented):
What problem is this trying to solve?

Sandychuang8:
The test hanging with infinite "Taking a while for active items to stop..." 
Although I don't see this issue that often anymore lately.

Yesterday I added this test, and I see that hanging error again, but with the latest LiteCore updates, I cannot repro it anymore...

` var config = CreateConfig(true, true, false);
            var replicator = new Replicator(config);

            LoadDocs();
            replicator.Start();
            LoadDocs();
            replicator.Start();
            replicator.Start();
            LoadDocs();
            replicator.Start();
            replicator.Start();
            LoadDocs();
            replicator.Start();`

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -246,12 +246,17 @@ public void Start(bool reset)
...
-                    ServerCertificate = null;
-                    WriteLog.To.Sync.I(Tag, $"{this}: Starting");
-                    Native.c4repl_start(_repl, Config.Options.Reset || reset);
-                    Config.Options.Reset = false;
-                    Config.Database.AddActiveStoppable(this);
                     status = Native.c4repl_getStatus(_repl);
+                    if (status.level == C4ReplicatorActivityLevel.Stopped
+                    || status.level == C4ReplicatorActivityLevel.Stopping
+                    || status.level == C4ReplicatorActivityLevel.Offline) {
+                        ServerCertificate = null;
```
Sandychuang8, line 253: But this change can prevent ServerCertificate set to null during replication and user call replicator Start().

borrrden (approved):

PR #1301, opened 2021-07-02
CBL-2094 Fix memory leak in WebSocketWrapper

Sandychuang8:
* Observing Reachability Changed event when Starting socket, and Stop observing when socket is closed.
* Reduce MaxReceivedBytesPending size
* Make sure socket is not closed before calling c4socket_gotHTTPResponse

### DIFF
--- a/src/Couchbase.Lite.Shared/Support/UWP/Reachability.cs
+++ b/src/Couchbase.Lite.Shared/Support/UWP/Reachability.cs
@@ -28,19 +28,19 @@ namespace Couchbase.Lite.Support
     [CouchbaseDependency(Transient = true)]
     internal sealed class Reachability : IReachability
     {
-#region Variables
+        #region Variables
 
         public event EventHandler<NetworkReachabilityChangeEventArgs> StatusChanged;
 
-#endregion
+        #endregion
         
-#region Properties
+        #region Properties
         
         public Uri Url { get; set; }
         
-#endregion
+        #endregion
 
-#region Private Methods
+        #region Private Methods
 
         private void OnNetworkStatusChanged(object sender)
         {
@@ -52,7 +52,7 @@ namespace Couchbase.Lite.Support
 			Task.Factory.StartNew(() => { StatusChanged?.Invoke(this, new NetworkReachabilityChangeEventArgs(status)); });
         }
 
-#endregion
+        #endregion
 
         #region IReachability
 
@@ -64,11 +64,6 @@ namespace Couchbase.Lite.Support
         public void Stop()
         {
             NetworkInformation.NetworkStatusChanged -= OnNetworkStatusChanged;
-        }
-        
-        public void Dispose()
-        {
-            Stop();
         }
 
         #endregion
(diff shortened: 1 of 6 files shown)


src/Couchbase.Lite.Shared/Support/iOS/iOSReachability.cs
```diff
@@ -96,9 +96,9 @@ private void NotifyFlagsChanged(NetworkReachabilityFlags flags)
             });
         }
 
-        #endregion
```
Sandychuang8, line 99: VS is automatically change the indentation when I just modify the code. I will change them back before merge this PR.

src/Couchbase.Lite.Shared/Support/UWP/Reachability.cs
```diff
@@ -65,6 +65,11 @@ public void Stop()
         {
             NetworkInformation.NetworkStatusChanged -= OnNetworkStatusChanged;
         }
+        
+        public void Dispose()
+        {
+            Stop();
+        }
```
Sandychuang8, line 72: Originally I want to set the Url to null in Dispose....

src/Couchbase.Lite.Shared/Support/iOS/iOSReachability.cs
borrrden: This is a setting you can modify

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -105,7 +105,7 @@ internal sealed class WebSocketWrapper
     {
         #region Constants
 
-        private const uint MaxReceivedBytesPending = 100 * 1024;
+        private const uint MaxReceivedBytesPending = 100 * 850;
```
borrrden, line 108: Is this still needed even with the other changes?  I don't really want to reduce this amount since it has an effect on performance of the network connection.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -295,6 +295,15 @@ private unsafe void Connected()
             });
         }
 
+        private void StopReachability()
+        {
+            if (_reachability != null)
```
borrrden, line 300: Also set to null here when done?

borrrden (changes requested):

src/Couchbase.Lite.Shared/Support/UWP/Reachability.cs
Sandychuang8: Remove Dispose method, just simply call Stop() instead since url needs to be existed as long as WebSocketWrapper/Replicator stays alive...

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
Sandychuang8: I guess it's a trade off between space vs speed.. Let me revert it.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
Sandychuang8: This _reachability is readonly in WebSocketWrapper...

borrrden: Have you observed what happens if `StopReachability` is called twice somehow?

borrrden (approved):

PR #1303, opened 2021-07-07
CBL-2102 Fix memory leak during MutableArrayObject.AddBlob while the Blob Content is large in size

Sandychuang8:
Revert using FLSlotSet in FleeceMutableArray to prevent memory leak when calling MutableArrayObject.AddBlob while the Blob Content is large in size.

### DIFF
--- a/src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
+++ b/src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
@@ -25,36 +25,27 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using Couchbase.Lite.Internal.Doc;
+using System.Diagnostics;
 
 namespace Couchbase.Lite.Fleece
 {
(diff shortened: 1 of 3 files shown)


borrrden:
Needs a better title

src/Couchbase.Lite.Shared/Serialization/FleeceMutableArray.cs
```diff
@@ -105,7 +97,7 @@ public void InitInSlot(MValue mv, MCollection parent)
         public void RemoveRange(int start, int count = 1)
         {
             if (!IsMutable) {
-                throw new InvalidOperationException(CouchbaseLiteErrorMessage.CannotRemoveItemsFromNonMutableMArray);
+                throw new InvalidOperationException("Cannot remove items from a non-mutable FleeceMutableArray");
```
borrrden, line 100: This change doesn't seem necessary

borrrden (changes requested):
I don't think it is necessary to revert all the error messages

borrrden (approved):
I think this is simply a revert to a previously known working version so I won't take too much more time analyzing it

PR #1304, opened 2021-07-09
CBL-2120 Retain c4socket when WebSocket starts and a release c4socket after c4socket is closed

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -564,6 +573,7 @@ private async void PerformWrite()
                         if (!_closed) {
                             unsafe {
                                 Native.c4socket_completedWrite(_socket, (ulong) nextData.Length);
+                                WriteLog.To.Sync.I(Tag, "c4Socket completed Write.");
```
borrrden, line 576: This one is going to get noisy so I think it should be left out, or moved to verbose.

borrrden (changes requested):

PR #1307, opened 2021-07-27
CBL-2186 Check for Network Task IsCompletedSuccessfully and IsFaulted 

Sandychuang8:
Try catch more host connection exceptions and handle them if any.


borrrden (approved):
Should be fine but has this made any observable difference?

Sandychuang8:
The observable is functioning fine as far as I can see.

PR #1308, opened 2021-07-29
CBL-1954 Use new NetworkErrorCode enum values

Sandychuang8:
Replace existing Posix error code with newly defined network error codes.
Remove PosixStatus class.
Update LiteCore to Commit: b3863874f33cd20895b88f3431dcf81538c9b25b [b386387]


Sandychuang8:
I don't recall I did anything to stop the PR validations from running??

src/LiteCore/test/LiteCore.Tests.Shared/DatabaseTest.cs
```diff
@@ -62,6 +67,11 @@ private void AssertMessage(C4ErrorDomain domain, int code, string expected)
             msg.Should().Be(expected, "because the error message should match the code");
         }
 
+        private int GetPosixExist(string name)
```
borrrden, line 70: Very strange function, if this is a code that is consistent between platforms, you can just use EEXIST directly in the test without even having this method.  The reason I recommended one was in case the error was different between platforms (also why pass in the name if you don't use it?)

src/LiteCore/src/LiteCore.Shared/Interop/C4Error_defs.cs
```diff
@@ -91,6 +91,17 @@ internal enum C4NetworkErrorCode : int
...
+        NetworkReset,
+        ConnectionAborted,
+        ConnectionReset,
+        ConnectionRefused,
+        NetworkDown,         // 20
+        NetworkUnreachable,
+        NotConnected,
+        HostDown,
+        HostUnreachable,
+        AddressNotAvailableAIL,
```
borrrden, line 103: Something weird happened here...I think this might be a binding generator bug?

src/Couchbase.Lite.Tests.Shared/URLEndpointListenerTest.cs
```diff
@@ -141,7 +141,7 @@ public void TestBusyPort()
             _listener = CreateListener(false);
             //listener1 uses the same port as listener
             var config = CreateListenerConfig(false);
-            var listener1 = Listen(config, PosixBase.GetCode(nameof(PosixWindows.EADDRINUSE)), CouchbaseLiteErrorType.POSIX);
+            var listener1 = Listen(config, (int)C4NetworkErrorCode.AddressNotAvailableAIL, CouchbaseLiteErrorType.POSIX);
```
borrrden, line 144: Wrong domain?

borrrden (changes requested):

src/Couchbase.Lite.Tests.Shared/URLEndpointListenerTest.cs
Sandychuang8: Currently there are 4 domains to choose from: CouchbaseLite, POSIX, SQLite, and Fleece. Are we adding new type such as Network?

borrrden: If this is on the public side of things, then currently network domain are translated to the CouchbaseLite domain with an error code offset.

src/Couchbase.Lite.Shared/Support/Status.cs
```diff
@@ -110,16 +139,21 @@ public static unsafe void ConvertNetworkError(Exception e, C4Error* outError)
...
                         message = ie.Message;
-                            c4err.domain = C4ErrorDomain.POSIXDomain;
-                            c4err.code = PosixBase.GetCode(nameof(PosixWindows.ECONNRESET));
+                            c4err.domain = C4ErrorDomain.NetworkDomain;
+                            c4err.code = (int)C4NetworkErrorCode.ConnectionReset;
                         }
-                    #endif
+// #elif __ANDROID__
+// EPIPE = .NET SocketException's SocketErrorCode ConnectionReset
+// Broken pipe is caused by the system sending SIGPIPE signal, which is an inter-process communication mechanism of Linux.
```
borrrden, line 154: What is this comment here for?

Sandychuang8: I think it's a good info. Let me move to a more proper location.

borrrden (changes requested):
The network error codes need to be added to the other public network error codes so that the user can actually make sense of them in the replicator callback.

borrrden (approved):

PR #1309, opened 2021-07-30
CBL-2191 Update HasAttachment flag if doc contains blob during SaveResolvedDocument

Sandychuang8:
Add the test to verify the C4DocumentFlag to contain the proper flags during and after conflict resolving.

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
@@ -1672,71 +1672,61 @@ namespace Test
         [Fact]
         public void TestConflictResolverReturningBlobWithFlagChecking()
         {
-            C4DocumentFlags flags= (C4DocumentFlags)0;;
-            //return new doc with a blob object
-            unsafe
+            C4DocumentFlags flags = (C4DocumentFlags)0;
+            unsafe 
             {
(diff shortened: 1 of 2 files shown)


src/Couchbase.Lite.Tests.Shared/ReplicationTest.cs
```diff
@@ -1669,6 +1669,44 @@ public void TestConflictResolverReturningBlob()
...
+            {
+                ((DictionaryObject) doc.GetValue("nestedBlob")).GetBlob("blob")?.Content.Should().ContainInOrder(new byte[] { 6, 6, 6 });
+                unsafe
+                {
+                    flags = doc.c4Doc.RawDoc->flags;
+                    flags.HasFlag(C4DocumentFlags.DocExists | C4DocumentFlags.DocHasAttachments).Should().BeTrue();
+                }
+            }
+        }
+
```
Sandychuang8, line 1709: @bmeike @jayahariv I added this test to verify `kHasAttachments flag may be lost in Conflict Resolver` and test has passed. Is my test correct to verify the issue?

Sandychuang8: Never mind @bmeike @jayahariv I actually found an issue with my test and now all is fixed.

borrrden (approved):

PR #1323, opened 2021-12-03
CBL-1946 Enhanced Configuration API in C# 9

Sandychuang8:
- Enhanced Configuration API in C# 9
- Add unit tests

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
@@ -77,7 +77,7 @@ namespace Couchbase.Lite.Sync
     /// <summary>
     /// A record representing configuration options for a <see cref="Replicator"/>
     /// </summary>
-    public sealed record ReplicatorConfiguration
+    public sealed partial record ReplicatorConfiguration
     {
         #region Constants
 
@@ -245,19 +245,6 @@ namespace Couchbase.Lite.Sync
         /// </summary>
         [CanBeNull]
         public IConflictResolver ConflictResolver { get; init; }
-
-        #if COUCHBASE_ENTERPRISE
-        /// <summary>
-        /// Get or set the way that the replicator will validate TLS certificates.  This
-        /// property will be overriden if the <see cref="PinnedServerCertificate"/> property
-        /// is set.
-        /// </summary>
-        public bool AcceptOnlySelfSignedServerCertificate
-        {
-            get => Options.AcceptOnlySelfSignedServerCertificate;
-            init => Options.AcceptOnlySelfSignedServerCertificate = value;
-        }
-        #endif
 
         internal TimeSpan CheckpointInterval
         {
--- a/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
@@ -73,7 +73,7 @@ namespace Couchbase.Lite.Sync
         AccessRemoved = 1 << 1
     }
 
-#if NET6_0_WINDOWS || NET6_0_WINDOWS10_0_17763_0 //Needs to specify TargetPlatform Version, just net6.0-windows/NET6_0_WINDOWS doesn't work as expected..
+#if NET6_0_OR_GREATER //|| NET6_0_WINDOWS || NET6_0_WINDOWS10_0_17763_0 //Needs to specify TargetPlatform Version, just net6.0-windows/NET6_0_WINDOWS doesn't work as expected..
     /// <summary>
     /// A record representing configuration options for a <see cref="Replicator"/>
     /// </summary>
--- a/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
@@ -21,6 +21,9 @@ using System.Collections.Generic;
 using System.Security.Cryptography.X509Certificates;
 
 using Couchbase.Lite.Internal.Logging;
+#if COUCHBASE_ENTERPRISE
+using Couchbase.Lite.P2P;
+#endif
 using Couchbase.Lite.Support;
 using Couchbase.Lite.Util;
 
@@ -87,9 +90,6 @@ namespace Couchbase.Lite.Sync
 
         #region Variables
 
-        [NotNull] private readonly Freezer _freezer = new Freezer();
-        private Database _otherDb;
-        private Uri _remoteUrl;
         private C4SocketFactory _socketFactory;
 
         #endregion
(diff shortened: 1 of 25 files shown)


src/Couchbase.Lite.Support.NetDesktop/Couchbase.Lite.Support.NetDesktop.csproj
```diff
@@ -5,8 +5,8 @@
   </PropertyGroup>
   <PropertyGroup>
     <Configurations>Debug;Release;Packaging;PackagingDebug</Configurations>
-    <TargetFramework Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp3.1</TargetFramework>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp3.1;net6.0-maccatalyst</TargetFrameworks>
```
borrrden, line 8: I have a feeling this is wrong, since it will add Mac Catalyst to builds on Linux as well.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -73,6 +73,244 @@ public enum DocumentFlags
...
+        {
+            get => Options.AcceptOnlySelfSignedServerCertificate;
+            init => Options.AcceptOnlySelfSignedServerCertificate = value;
+        }
+        #endif
+
+        internal TimeSpan CheckpointInterval
+        {
+            get => Options.CheckpointInterval;
+            set => _freezer.PerformAction(() => Options.CheckpointInterval = value);
```
borrrden, line 265: These can all be replaced with `init` too, and the freezer removed

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -73,6 +73,244 @@ public enum DocumentFlags
...
+        [CanBeNull]
+        public IConflictResolver ConflictResolver { get; init; }
+
+        #if COUCHBASE_ENTERPRISE
+        /// <summary>
+        /// Get or set the way that the replicator will validate TLS certificates.  This
+        /// property will be overriden if the <see cref="PinnedServerCertificate"/> property
+        /// is set.
+        /// </summary>
+        public bool AcceptOnlySelfSignedServerCertificate
```
borrrden, line 255: Why has this moved here instead of staying in the EE repo?

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
Sandychuang8: Actually record can be partial as well.. I can do either way. Maybe I should make it consistent as class?

borrrden: It will make the diff on the EE side easier to read if it remains there, plus the CE repo won't be cluttered with more `#if`

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -21,6 +21,9 @@
 using System.Security.Cryptography.X509Certificates;
 
 using Couchbase.Lite.Internal.Logging;
+#if COUCHBASE_ENTERPRISE
+using Couchbase.Lite.P2P;
+#endif
```
borrrden, line 26: What's this for?

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -73,6 +76,230 @@ public enum DocumentFlags
         AccessRemoved = 1 << 1
     }
 
+#if NET6_0_WINDOWS || NET6_0_WINDOWS10_0_17763_0 //Needs to specify TargetPlatform Version, just net6.0-windows/NET6_0_WINDOWS doesn't work as expected..
```
borrrden, line 79: Maybe you should define the former manually, since the latter is not a guarantee, especially on our internal build machines (are you planning to get an upgrade to 2022 internally?)

src/Couchbase.Lite.Shared/API/Sync/URLEndpoint.cs
```diff
@@ -83,7 +83,9 @@ public URLEndpoint([NotNull]Uri url)
 
         void IEndpointInternal.Visit(ReplicatorConfiguration config)
         {
+            #if !NET6_0_WINDOWS10_0_17763_0
             config.RemoteUrl = Url;
```
borrrden, line 87: How is this getting handled in the .NET 6 case?

Sandychuang8: Handled in the ReplicatorConfiguration record constructor.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
Sandychuang8: FYI (https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#design): So complicated.. Will need additional time to figure this out to apply to our project..

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
Sandychuang8: Need to use this to check MessageEndpoint Type: https://github.com/couchbase/couchbase-lite-net/pull/1323/files#diff-825d20c5c8cd1686646490f8995630790384cc7ff342c9e1a1a66f481c986a19R292-R296

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
borrrden: That's basically why I don't see the need to rush to put in .NET 6.  Why not wait until we can deprecate the older runtimes?

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -73,6 +76,230 @@ public enum DocumentFlags
...
+                RemoteUrl = ((URLEndpoint)target).Url;
+            }
+            #if COUCHBASE_ENTERPRISE
+            else if(target.GetType() == typeof(DatabaseEndpoint)) {
+                OtherDB = ((DatabaseEndpoint)target).Database;
+            } else if (target.GetType() == typeof(MessageEndpoint)) {
+                _socketFactory = ((MessageEndpoint)target).SocketFactory;
+                RemoteUrl = new Uri("x-msg-endpoint://");
+                Options.RemoteDBUniqueID = ((MessageEndpoint)target).Uid;
+            }
```
borrrden, line 296: If this is what it is going to take to make things work with internal init-only properties than I changed my mind and the old way is better.  This is leaking too much EE implementation stuff into here.

Sandychuang8: Agree, and the old way looks much cleaner. Reverted.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
Sandychuang8: Created CBL-2631

borrrden:
Will revisit this in 4.0

PR #1329, opened 2022-01-20
CBL-2700 CBL-2612 LiveQuerier and tests

Sandychuang8:


Sandychuang8:
Something has changed in LiteCore libraries restore in Github Actions. Will take a look and fix it in next sprint.

src/Couchbase.Lite.Shared/API/Query/Parameters.cs
```diff
@@ -210,7 +220,9 @@ public Parameters SetValue([NotNull]string name, [CanBeNull]object value)
             }
 
             _freezer.PerformAction(() => _params[name] = value);
-            
+            if (_query != null)
+                _query.SetParameters(this.ToString());
```
borrrden, line 224: Modernize: `_query?.SetParameters` ?

src/Couchbase.Lite.Shared/Query/LiveQuerier.cs
```diff
@@ -0,0 +1,177 @@
...
+        private const string Tag = nameof(LiveQuerier);
+
+        private static readonly C4QueryObserverCallback QueryCallback = QueryObserverCallback;
+
+        #endregion
+
+        #region Variables
+
+        [NotNull] private readonly Event<QueryChangedEventArgs> _changed = new Event<QueryChangedEventArgs>();
+        private C4QueryObserver* _queryObserver; //uniqe long value
```
borrrden, line 43: What is this comment saying?

src/Couchbase.Lite.Shared/Query/QueryBase.cs
```diff
@@ -89,108 +117,76 @@ public QueryBase()
...
-            {
-                return;
-            }
-
-            // External updates should poll less frequently
-            var updateInterval = _updateInterval;
+            if (_c4Query == null) {
+                C4Query* query = (C4Query*)ThreadSafety.DoLockedBridge(err =>
+                {
+                    if (this is XQuery) {
```
borrrden, line 141: When this starts happening it's an indication that you've put too much into your base class.  It's now so much that you need to resort to runtime type checking to figure out what kind of query it is.  I would recommend pushing this functionality down to the subclasses (NQuery vs XQuery).

src/Couchbase.Lite.Shared/Query/QueryBase.cs
```diff
@@ -68,17 +65,48 @@ public Parameters Parameters
...
+
+        internal SerialQueue DispatchQueue { get; } = new SerialQueue();
+
+        #endregion
+
+        #region Properties - Json Query Expression
+
+        protected bool Distinct { get; set; }
+
+        protected QueryDataSource FromImpl { get; set; }
```
borrrden, line 85: Why do all these need to be here?  These only apply to XQuery anyway...the point of having different classes becomes lost at this point.

borrrden (changes requested):
In my opinion, this should be refactored to not rely so much on everything being in QueryBase with `if(this is XQuery)` sprinkled everywhere.

src/Couchbase.Lite.Shared/Query/QueryBase.cs
```diff
@@ -243,12 +185,35 @@ public ListenerToken AddChangeListener([NotNull] EventHandler<QueryChangedEventA
...
+        }
+
+        #endregion
+
+        #region QueryBase
 
-            _willUpdate.Set(false);
+        protected abstract void CreateQuery();
+
+        internal abstract unsafe Dictionary<string, int> CreateColumnNames(C4Query* query);
```
borrrden, line 201: Can you point out the need to make this an internal method?  What is the benefit of moving it here?  If none, it's better to keep it encapsulated like where it was before.  I don't see QueryBase calling it, or any other class calling it on QueryBase.

Sandychuang8: It is used to get QueryResultSet in LiveQuerier class. Called in QueryObserverCalled(C4QueryObserver* obs, C4Query* query) method.

borrrden: I think this logic is still breaking outside of the class too much.  If you want to make sure ColumnNames is not calculated when you get it from inside of QueryBase, then how about just having a method called `ColumnNames()` that returns it directly, and calculates it if needed?

borrrden (approved):

PR #1332, opened 2022-02-12
CBL-2810 Mapping Android undefined Exceptions into CBL errors

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -573,6 +573,7 @@ private async void PerformWrite()
                         await stream.WriteAsync(nextData, 0, nextData.Length, cancelSource.Token).ConfigureAwait(false);
                     } catch (Exception e) {
                         DidClose(e);
+                        await NetworkStream?.FlushAsync(cancelSource.Token);
```
borrrden, line 576: 575 is going to set `NetworkStream` to `null` so it seems like this line is not doing anything.

borrrden (changes requested):

borrrden (approved):

borrrden:
No longer mergeable and motivation unclear

PR #1334, opened 2022-03-08
CBL-2897 Update Websocket close process in correct sequence

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -178,7 +177,7 @@ public unsafe void CloseSocket()
                     WriteLog.To.Sync.I(Tag, "Closing socket normally due to request from LiteCore");
                     ReleaseSocket(new C4Error(0, 0));
                 });
-            });
+            }).ContinueWith(t=> ResetConnections());
```
pasin, line 180: Sorry, I don't know .NET. Does this get executed before or after `ReleaseSocket(new C4Error(0, 0));`?

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -708,6 +703,11 @@ private void ResetConnections()
                 lock (_writeQueueLock) {
                     Misc.SafeSwap(ref _writeQueue, null);
                 }
+
+                NetworkStream?.Dispose();
+                NetworkStream = null;
+                _readWriteCancellationTokenSource?.Dispose();
+                _readWriteCancellationTokenSource = null;
```
pasin, line 710: Should the following be done after waiting for the write queue as well?

```
_client?.Dispose();
_client = null;
_readWriteCancellationTokenSource?.Cancel();

_receivePause?.Dispose();
_receivePause = null;
```

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
pasin: Reset first and then release Socket.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
Sandychuang8: _client should be disposed asap, and _readWriteCancellationTokenSource need to call cancel soon as well, and also we don't want to receive anything at this point so dispose it should be fine. I think only things matter to wait for finishing the write queue are NetworkStream and _readWriteCancellationTokenSource..

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
Sandychuang8: Let me reverse the order.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
pasin: Sorry, I am not familiar with all the code here so I will need to ask questions.

If disposing `_client` and `_readWriteCancellationTokenSource`, can anything in the write queue be sent?

Sandychuang8: My original thought is to stop write/sending to remote at this point since we are not going to care the responses but now I see your point. Let me make corrections... The goal is to keep the connection and finish sending all the write queue to remote before socket close.

pasin: Here is the comment about the close function.

https://github.com/couchbase/couchbase-lite-core/blob/master/C/include/c4SocketTypes.h#L113-L120

CC: @borrrden for review

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
borrrden: I thought @pasin said `ResetConnections` should be last....if it needs to happen *before* the logic inside of the c4Queue then a different approach is needed.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
borrrden: I agree that if we are waiting for the writeQueue to be "complete" we need to empty it, but it will never be empty unless we process all the messages out of it.  It makes the wait loop on writeQueue pointless since if the queue is not empty, waiting is not going to change that.  That's the smell for me.  If we want to ignore any further writes then we don't need to wait for the writeQueue to be completed.

PR #1364, opened 2022-05-05
CBL-3058 Scopes and Collections Prototype

Sandychuang8:
- Database, Document, Query
- Replicator Configuration
- ListenerToken Remove() - [CBL-1762](https://issues.couchbase.com/browse/CBL-1762)
***** Not all API comments are updated *****

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -1,0 +1,116 @@
+﻿using Couchbase.Lite.Internal.Query;
+using Couchbase.Lite.Support;
+using JetBrains.Annotations;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Couchbase.Lite
+{
+    public sealed class Collection : ICollection, IDisposable, IChangeObservable<DatabaseChangedEventArgs>
+    {
+        [NotNull]
+        internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
+
+        public string Name => throw new NotImplementedException();
+
+        public Scope Scope => throw new NotImplementedException();
+
+        public ulong Count => throw new NotImplementedException();
+
+        public ListenerToken<DatabaseChangedEventArgs> AddChangeListener([CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DatabaseChangedEventArgs> handler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public ListenerToken<DatabaseChangedEventArgs> AddChangeListener([NotNull] EventHandler<DatabaseChangedEventArgs> handler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public ListenerToken<DocumentChangedEventArgs> AddDocumentChangeListener([NotNull] string id, [CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public ListenerToken<DocumentChangedEventArgs> AddDocumentChangeListener([NotNull] string id, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void CreateIndex([NotNull] string name, [NotNull] IndexConfiguration indexConfig)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Delete([NotNull] Document document)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool Delete([NotNull] Document document, ConcurrencyControl concurrencyControl)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void DeleteIndex([NotNull] string name)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Dispose()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Document GetDocument([NotNull] string id)
+        {
+            throw new NotImplementedException();
+        }
+
+        public DateTimeOffset? GetDocumentExpiration(string docId)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IList<string> GetIndexes()
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Purge([NotNull] Document document)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Purge([NotNull] string docId)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void RemoveChangeListener(ListenerToken<DatabaseChangedEventArgs> token)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Save([NotNull] MutableDocument document)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool Save([NotNull] MutableDocument document, ConcurrencyControl concurrencyControl)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool Save(MutableDocument document, Func<MutableDocument, Document, bool> conflictHandler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool SetDocumentExpiration(string docId, DateTimeOffset? expiration)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -15,7 +15,7 @@ namespace Couchbase.Lite
 
         public string Name => throw new NotImplementedException();
 
-        public Scope Scope => throw new NotImplementedException();
+        public IScope Scope => throw new NotImplementedException();
 
         public ulong Count => throw new NotImplementedException();
 
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -7,7 +7,8 @@ using System.Threading.Tasks;
 
 namespace Couchbase.Lite
 {
-    public sealed class Collection : ICollection, IDisposable, IChangeObservable<DatabaseChangedEventArgs>
+    public sealed class Collection : ICollection, IDisposable, IChangeObservable<DatabaseChangedEventArgs>,
+        IDocumentChangeObservable
     {
         [NotNull]
         internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
@@ -18,22 +19,22 @@ namespace Couchbase.Lite
 
         public ulong Count => throw new NotImplementedException();
 
-        public ListenerToken<DatabaseChangedEventArgs> AddChangeListener([CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DatabaseChangedEventArgs> handler)
+        public ListenerToken AddChangeListener([CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DatabaseChangedEventArgs> handler)
         {
             throw new NotImplementedException();
         }
 
-        public ListenerToken<DatabaseChangedEventArgs> AddChangeListener([NotNull] EventHandler<DatabaseChangedEventArgs> handler)
+        public ListenerToken AddChangeListener([NotNull] EventHandler<DatabaseChangedEventArgs> handler)
         {
             throw new NotImplementedException();
         }
 
-        public ListenerToken<DocumentChangedEventArgs> AddDocumentChangeListener([NotNull] string id, [CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
+        public ListenerToken AddDocumentChangeListener([NotNull] string id, [CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
         {
             throw new NotImplementedException();
         }
 
-        public ListenerToken<DocumentChangedEventArgs> AddDocumentChangeListener([NotNull] string id, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
+        public ListenerToken AddDocumentChangeListener([NotNull] string id, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
         {
             throw new NotImplementedException();
         }
@@ -88,7 +89,7 @@ namespace Couchbase.Lite
             throw new NotImplementedException();
         }
 
-        public void RemoveChangeListener(ListenerToken<DatabaseChangedEventArgs> token)
+        public void RemoveChangeListener(ListenerToken token)
         {
             throw new NotImplementedException();
         }
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -1,4 +1,22 @@
-﻿using Couchbase.Lite.Internal.Query;
+﻿// 
+//  Collection.cs
+// 
+//  Copyright (c) 2022 Couchbase, Inc All rights reserved.
+// 
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+// 
+//  http://www.apache.org/licenses/LICENSE-2.0
+// 
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// 
+
+using Couchbase.Lite.Internal.Query;
 using Couchbase.Lite.Support;
 using JetBrains.Annotations;
 using System;
@@ -10,14 +28,20 @@ namespace Couchbase.Lite
     public sealed class Collection : ICollection, IDisposable, IChangeObservable<DatabaseChangedEventArgs>,
         IDocumentChangeObservable
     {
-        [NotNull]
-        internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
+        #region Properties
+
+        [NotNull] //TODO: expose Database.ThreadSafety and assign here
+        internal ThreadSafety ThreadSafety { get; }
 
         public string Name => throw new NotImplementedException();
 
         public IScope Scope => throw new NotImplementedException();
 
         public ulong Count => throw new NotImplementedException();
+
+        #endregion
+
+        #region IChangeObservable
 
         public ListenerToken AddChangeListener([CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DatabaseChangedEventArgs> handler)
         {
@@ -29,6 +53,10 @@ namespace Couchbase.Lite
             throw new NotImplementedException();
         }
 
+        #endregion
+
+        #region IDocumentChangeObservable
+
         public ListenerToken AddDocumentChangeListener([NotNull] string id, [CanBeNull] TaskScheduler scheduler, [NotNull] EventHandler<DocumentChangedEventArgs> handler)
         {
             throw new NotImplementedException();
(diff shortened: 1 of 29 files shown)


src/Couchbase.Lite.Shared/API/Database/ListenerToken.cs
```diff
@@ -19,34 +19,53 @@
...
             Type = type;
+            Event = e;
         }
 
         #endregion
+
+        public int Remove()
+        {
+            return Event.Remove(this);
+        }
```
Sandychuang8, line 69: @borrrden Can you advise the elegant/best way you can think of to implement ListenToken.Remove()?

borrrden: Can you make the token aware of its source (replication, database, etc) and then just call the existing method?

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -0,0 +1,116 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Couchbase.Lite
+{
+    public sealed class Collection : ICollection, IDisposable, IChangeObservable<DatabaseChangedEventArgs>
+    {
+        [NotNull]
+        internal ThreadSafety ThreadSafety { get; } = new ThreadSafety();
```
borrrden, line 13: I don't think this will be needed.  Collection can use the database threadsafety.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public string Name => throw new NotImplementedException();
+
+        public Scope Scope => throw new NotImplementedException();
```
borrrden, line 17: IScope?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -126,6 +127,9 @@ public sealed unsafe partial class Database : IDisposable
         private static readonly C4DocumentObserverCallback _DocumentObserverCallback = DocObserverCallback;
         private static readonly C4DatabaseObserverCallback _DatabaseObserverCallback = DbObserverCallback;
 
+        public const string DefaultScopeName = "_default";
+        public const string DefaultCollectionName = "_default";
```
borrrden, line 131: Don't use public const, it is bad for API since the value is copied into the consuming assembly (i.e. not referencing it at runtime).  It means that any change would be an API break.  If it is public, use `static readonly`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -170,7 +174,8 @@ public sealed unsafe partial class Database : IDisposable
         public DatabaseConfiguration Config { get; }
 
         /// <summary>
-        /// Gets the number of documents in the database
+        /// [Obsolete("Count is deprecated, please use <see cref="GetDefaultCollection().Count"/>.")]
```
borrrden, line 177: Why is this commented?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -373,6 +380,93 @@ internal Database(C4Database* c4db)
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Get scope names that have at least one collection.
+        /// Note: the default scope is exceptional as it will always be listed even though there are no collections    
+        /// under it.
+        /// </summary>
+        /// <returns></returns>
+        public List<Scope> GetScopes()
```
borrrden, line 409: Prefer interfaces for API definition (IReadOnlyList)

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -633,9 +770,9 @@ public void CreateIndex([@NotNull] string name, [@NotNull] IndexConfiguration in
             });
         }
 
-        /// Creates a Query object from the given N1QL query string.
+        /// Creates a Query object from the given SQL string.
```
borrrden, line 773: SQL++

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -373,6 +380,93 @@ internal Database(C4Database* c4db)
...
+        public Scope GetScope(string name)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Get all collections of given Scope name.
+        /// </summary>
+        /// <returns></returns>
+        public List<Collection> GetCollections(string scope = DefaultScopeName)
```
borrrden, line 430: Prefer interfaces -> IReadOnlyList\<ICollection>

src/Couchbase.Lite.Shared/API/Database/Scope.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Couchbase.Lite.Query;
+using JetBrains.Annotations;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Couchbase.Lite
+{
+    public sealed class Scope : IScope, IDisposable, IChangeObservable<DatabaseChangedEventArgs>
```
borrrden, line 9: Probably doesn't need to be IDisposable

src/Couchbase.Lite.Shared/API/Query/IQuery.cs
```diff
@@ -79,29 +79,29 @@ public interface IQuery : IDisposable
...
+        ///// <returns>A token that can be used to remove the listener later</returns>
+        //ListenerToken<QueryChangedEventArgs> AddChangeListener([NotNull] EventHandler<QueryChangedEventArgs> handler);
 
         /// <summary>
         /// Removes a changes listener based on the token that was received from
         /// <see cref="AddChangeListener(TaskScheduler, EventHandler{QueryChangedEventArgs})"/>
         /// </summary>
         /// <param name="token">The received token from adding the change listener</param>
-        void RemoveChangeListener([NotNull]ListenerToken token);
+        //void RemoveChangeListener([NotNull]ListenerToken token);
```
borrrden, line 104: Remove these instead of commenting

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -346,6 +370,33 @@ public ReplicatorConfiguration([NotNull] Database database, [NotNull] IEndpoint
...
+
+        public void RemoveCollection(Collection collection)
+        {
+            CollectionConfigs.Remove(collection);
+        }
+
+        [CanBeNull]
+        public CollectionConfiguration GetCollectionConfig(Collection collection)
+        {
+            return CollectionConfigs[collection];
```
borrrden, line 395: This will throw an exception if collection doesn't exist.  Use TryGet

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
```diff
@@ -71,24 +71,39 @@ public struct ReplicatedDocument
...
+
         internal bool IsTransient { get; }
 
         internal C4Error NativeError { get; }
 
-        internal ReplicatedDocument([NotNull]string docID, C4RevisionFlags flags, C4Error error,
+        internal ReplicatedDocument([NotNull] string docID, C4RevisionFlags flags, C4Error error,
+            bool isTransient)
+        {
+            Collection = null;
```
borrrden, line 83: It would be ideal if this were not allowed, and instead the default collection could be passed in.

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Because of the below spec, 
_The Collection(Collection) function is added and the Database(Database) factory function will be deprecated in DataSource._
method returns DatabaseSource(collection, ThreadSafety) and there is no where to access database threadsafety.. Did I miss anything?

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: If the default value is static readonly and I use the default value as public IReadOnlyList<ICollection> GetCollections(string scope = DefaultScopeName), I will be getting _Default parameter for value must be a compile time constant_..

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: ? Count in Database is marked deprecated in spec. And Obsolete comments will direct user to use the alternative way to get the Count. Did I miss anything?

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -413,10 +415,13 @@ private static void OnDocEnded(C4Replicator* repl, bool pushing, IntPtr numDocs,
...
+            var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
+
             var replicatedDocumentsContainConflict = new List<ReplicatedDocument>();
             var documentReplications = new List<ReplicatedDocument>();
             for (int i = 0; i < (int) numDocs; i++) {
                 var current = docs[i];
+                //var collection = replicator.Config.Database.GetCollection(current->collectionName.CreateString());
                 if (!pushing && current->error.domain == C4ErrorDomain.LiteCoreDomain &&
                     current->error.code == (int) C4ErrorCode.Conflict) {
                     replicatedDocumentsContainConflict.Add(new ReplicatedDocument(current->docID.CreateString() ?? "",
```
Sandychuang8, line 427: About the question of setting Collection = null in ReplicatedDocument constructor, the collection is supposedly to be created and pass in ReplicatedDocument (internal and only is used here) as parameter. The old constructor will be removed and setting Collection = null is temporarily because Database.GetCollection wasn't implemented yet, and I still want to existing tests to pass until the new implementation goes in...

src/Couchbase.Lite.Shared/API/Database/Collection.cs
borrrden: What I mean is, you could just expose Database.ThreadSafety here from this property instead of creating a new one.

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: If you want to use a constant, you can have a private const, then have the public static readonly be the value of that const.

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: I mean, you have commented out the obsolete attribute, so it won't actually function like it should (giving compiler warning)

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: If it is temporary then ok

src/Couchbase.Lite.Shared/API/Document/Document.cs
```diff
@@ -96,6 +96,12 @@ internal C4DocumentWrapper c4Doc
         [CanBeNull]
         internal Database Database { get; set; }
 
+        /// <summary>
+        /// Gets the Collection that this document belongs to, if any
+        /// </summary>
+        [CanBeNull]
+        public Collection Collection { get; set; }
```
borrrden, line 103: ICollection (I didn't make an exhaustive list, but if we have these interfaces anyway let's use them)

src/Couchbase.Lite.Shared/API/Query/IIndexable.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public interface IIndexable
+    {
+        /// <summary>
+        /// Gets a list of index names that are present in the database
+        /// </summary>
+        /// <returns>The list of created index names</returns>
+        [NotNull]
+        [ItemNotNull]
+        IList<string> GetIndexes();
```
borrrden, line 17: Probably can be read only list

src/Couchbase.Lite.Tests.Shared/NotificationTest.cs
```diff
@@ -150,8 +150,8 @@ public async Task TestRemoveDocumentChangeListener()
             doc1.SetString("name", "Scott Tiger");
             Db.Save(doc1);
             _wa.WaitForResult(TimeSpan.FromSeconds(5));
-
-            Db.RemoveChangeListener(token);
+            token.Remove();
+            //Db.RemoveChangeListener(token);
```
borrrden, line 154: No need to leave in these commented lines throughout the tests

src/Couchbase.Lite.Shared/Query/LiveQuerier.cs
```diff
@@ -64,9 +64,8 @@ internal LiveQuerier(QueryBase queryBase)
 
         #region Internal Methods
 
-        internal unsafe LiveQuerier CreateLiveQuerier(C4Query* c4Query, ListenerToken listenerToken)
+        internal unsafe LiveQuerier CreateLiveQuerier(C4Query* c4Query)
         {
-            _listenerToken = listenerToken;
```
borrrden, line 69: Looks like _listenerToken is no longer used, so remove?

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: woops... Will move them out of comments..

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -301,6 +310,8 @@ private bool IsReadyToClose
             }
         }
 
+        public IScope Scope => GetDefaultScope();
```
borrrden, line 313: Implemented interface methods can use the inheritdoc tag for the doc comments

src/Couchbase.Lite.Shared/API/Database/Interfaces/IScope.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Gets the Scope Name
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Gets all collections in the Scope
+        /// </summary>
+        IReadOnlyList<Collection> Collections { get; }
```
borrrden, line 16: IReadOnlyList<ICollection>

src/Couchbase.Lite.Shared/API/Query/DataSource.cs
```diff
@@ -39,10 +40,25 @@ public static class DataSource
 
         /// <summary>
         /// Creates a data source for an <see cref="IQuery" /> that gets results from the given
+        /// <see cref="Collection(Lite.Collection)" />
+        /// </summary>
+        /// <param name="collection">The collection to operate on</param>
+        /// <returns>The source of data for the <see cref="IQuery" /></returns>
+        [NotNull]
+        public static IDataSourceAs Collection([NotNull] Collection collection)
```
borrrden, line 48: ICollection

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -101,6 +101,9 @@ public sealed partial class ReplicatorConfiguration
 
         #region Properties
 
+        [CanBeNull]
+        public IDictionary<Collection, CollectionConfiguration> CollectionConfigs { get; set; }
```
borrrden, line 105: ICollection

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
```diff
@@ -71,24 +71,39 @@ public struct ReplicatedDocument
         [CanBeNull]
         public CouchbaseException Error { get; internal set; }
 
+        public Collection Collection { get; }
```
borrrden, line 74: ICollection

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: aware of this already, but I haven't added comments in interfaces. comments will be polished during scopes and collections implementations.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -373,6 +384,93 @@ internal Database(C4Database* c4db)
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Create a named collection in the specified scope.
+        /// If the collection already exists, the existing collection will be returned.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public Collection CreateCollection(string name, string scope = _defaultScopeName)
```
borrrden, line 445: ICollection

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -731,12 +746,13 @@ public Document GetDocument([@NotNull]string id)
...
         /// <summary>
-        /// Gets a list of index names that are present in the database
+        /// [DEPRECATED] Gets a list of index names that are present in the database
         /// </summary>
         /// <returns>The list of created index names</returns>
+        [Obsolete("GetIndexes is deprecated, please use GetDefaultCollection().GetIndexes")]
         [@NotNull]
         [@ItemNotNull]
-        public IList<string> GetIndexes()
+        public IReadOnlyList<string> GetIndexes()
```
borrrden, line 755: I don't think you can change this, it would be considered breaking from the existing API right?

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -346,6 +370,39 @@ public ReplicatorConfiguration([NotNull] Database database, [NotNull] IEndpoint
 
         #endregion
 
+        #region Public Methods - Scopes and Collections
+
+        public void AddCollections(IList<Collection> collections, [CanBeNull]CollectionConfiguration config = null)
```
borrrden, line 375: ICollection

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -346,6 +370,39 @@ public ReplicatorConfiguration([NotNull] Database database, [NotNull] IEndpoint
...
+        {
+            foreach(var col in collections) {
+                CollectionConfigs.Add(col, config);
+            }
+        }
+
+        public void AddCollection(Collection collection, [CanBeNull] CollectionConfiguration config = null)
```
borrrden, line 382: ICollection

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -346,6 +370,39 @@ public ReplicatorConfiguration([NotNull] Database database, [NotNull] IEndpoint
...
+        {
+            CollectionConfigs.Add(collection, config);
+        }
+
+        public void RemoveCollection(Collection collection)
```
borrrden, line 387: ICollection

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -346,6 +370,39 @@ public ReplicatorConfiguration([NotNull] Database database, [NotNull] IEndpoint
...
+        {
+            CollectionConfigs.Remove(collection);
+        }
+
+        [CanBeNull]
+        public CollectionConfiguration GetCollectionConfig(Collection collection)
```
borrrden, line 393: ICollection

borrrden (changes requested):

borrrden (changes requested):
There appears to still be some open items leftover from the last review

src/Couchbase.Lite.Shared/API/Query/IIndexable.cs
Sandychuang8: Database class's GetIndexes() is also implementing this interface and if changes it will break API.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorProgress.cs
Sandychuang8: I think we already discussed this. After implementation started, this constructor will be removed.

src/Couchbase.Lite.Shared/API/Query/DataSource.cs
Sandychuang8: Please see below TODO commentsfrom line 51 to line 53

borrrden: I don't see any todo comments in this file

borrrden (approved):

PR #1366, opened 2022-05-11
CBL-2971 Add NetwrokInterface property in Replicator Config

Sandychuang8:


Sandychuang8:
Close PR because the implementation isn't completed.

PR #1370, opened 2022-05-15
CBL-2971 Add Network Interface property in Replicator Configuration 

Sandychuang8:


src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
```diff
@@ -175,8 +176,14 @@ public string Filter
         /// by a <see cref="Replicator"/> with this configuration will be rejected.
         /// </summary>
         [CanBeNull]
-        public X509Certificate2 PinnedServerCertificate { get; set; }
+        internal X509Certificate2 PinnedServerCertificate { get; set; }
```
borrrden, line 179: Why did this change to internal?

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -215,21 +216,43 @@ public unsafe void Start()
...
-                    _client.Client.DualMode = true;
-                } catch (ArgumentException) {
-                    WriteLog.To.Sync.I(Tag, "IPv4/IPv6 dual mode not supported on this device, falling back to IPv4");
-                    _client = new TcpClient(AddressFamily.InterNetwork);
+                if (!String.IsNullOrEmpty(_options.NetworkInterface)) {
+                    try {
+                        var localAddress = GetLocalNIAddress(_options.NetworkInterface);
+                        if (localAddress == null) {
+                            WriteLog.To.Sync.I(Tag, $"Unknown Network Interface {_options.NetworkInterface}.");
+                            DidClose(new CouchbaseNetworkException(C4NetworkErrorCode.UnknownHost));
```
borrrden, line 228: I don't think this error is appropriate.  Did you find it in another platform implementation?

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -215,21 +216,43 @@ public unsafe void Start()
...
+                            return;
+                        }
+
+                        IPEndPoint localEndPoint = new IPEndPoint(localAddress, 0);
+                        var socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+                        socket.Bind(localEndPoint);
+                        _client = new TcpClient() { Client = socket };
```
borrrden, line 235: TcpClient looks like it can take IPEndPoint inside the constructor, which I think would be more appropriate.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -264,6 +287,23 @@ public void Write(byte[] data)
...
 
+        internal IPAddress GetLocalNIAddress(string rni)
+        {
+            IPAddress addr = null;
+            if (IPAddress.TryParse(rni, out addr))
+                return addr;
+
+            foreach (NetworkInterface ni in NetworkInterface.GetAllNetworkInterfaces()) {
+                if (ni.Name == rni) { // UnicastAddresses[1] will give ipv4 address of certain adapter
+                    addr = ni.GetIPProperties().UnicastAddresses[0].Address; //This will give ipv6 address of certain adapter
```
borrrden, line 298: I worry that the IPv6 address may be non functional.  I think that it should at least be checked to make sure it is not [link local](https://docs.microsoft.com/en-us/dotnet/api/system.net.ipaddress.isipv6linklocal?view=net-6.0#system-net-ipaddress-isipv6linklocal)

borrrden (changes requested):

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
Sandychuang8: There is a public facing API in Replicator config, so I mark this one internal. (I know it will not make any difference since the class itself is internal.) I can change it back. But I will keep NetworkInterface here as internal...

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
Sandychuang8: Pasin said to match URLEndpointListener Config Network Interface invalid error. And base on the test https://github.com/couchbase/couchbase-lite-net/blob/master/src/Couchbase.Lite.Tests.Shared/URLEndpointListenerTest.cs#L615
new CouchbaseNetworkException(C4NetworkErrorCode.UnknownHost) is the exception with error. Please suggest a better one if you think it makes more sense.

borrrden: If the other one uses it then it's probably fine.  It's not likely to happen that often anyway.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
borrrden: I would guess that the answer is yes.  The latter might be doing some needless internal processing to find an appropriate address first in the parameterless constructor.

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
borrrden: It's mostly a style choice then, so either way is fine.

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -264,6 +276,61 @@ public void Write(byte[] data)
...
+
+            // If host is IP, use the family from the IP.
+            isRemoteHostIP = IPAddress.TryParse(_logic.UrlRequest.Host, out hostAddr);
+            if (isRemoteHostIP)
+                isRemoteHostAddrIPv6 = hostAddr.AddressFamily == AddressFamily.InterNetworkV6;
+
+            try {
+                //Create tcp client with ipv6 address for adapter
+                var ipv6 = localAddresses.FirstOrDefault(x => x.Address.AddressFamily == AddressFamily.InterNetworkV6)?.Address;
+                return new TcpClient(new IPEndPoint(ipv6, 0));
```
borrrden, line 320: This is not a good enough check.  There are three cases:  

1. Remote Host is an IPv6 address (handled) 
2. Remote Host is a DNS name (handled)
3. Remote Host is an IPv4 address (not handled)

The IP address checks should be done first.  Handle both IPv4 and IPv6 cases, and then fall back in the case of DNS name to either IPv4 or IPv6 (IPv4 safer).

borrrden (changes requested):

src/Couchbase.Lite.Shared/Sync/WebSocketWrapper.cs
```diff
@@ -264,6 +276,68 @@ public void Write(byte[] data)
...
+                    return GetTcpClient(AddressFamily.InterNetworkV6, addresses);
+                } else if (ni.Supports(NetworkInterfaceComponent.IPv4)) {
+                    return GetTcpClient(AddressFamily.InterNetwork, addresses);
+                }
+
+                WriteLog.To.Sync.I(Tag, $"Network Interface {_options.NetworkInterface} supports neither IPv6 nor IPv4.");
+                throw new CouchbaseNetworkException(C4NetworkErrorCode.UnknownHost);
+
+            } catch (Exception ex) {
+                throw ex;
```
borrrden, line 323: This part does nothing except what would happen without the catch.  Either make it do something else or just don't include the try catch.

borrrden (approved):

PR #1373, opened 2022-05-19
CBL-3053 Update the binding script to work with source based litecore exporting

Sandychuang8:


Sandychuang8:
Not correct, close pr.

PR #1381, opened 2022-06-04
CBL-3078 Add Scopes Collections management in Database

Sandychuang8:


src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -87,78 +147,226 @@ public void RemoveChangeListener(ListenerToken token)
...
+        {
+            using (var collName_ = new C4String(Name))
+            using (var scopeName_ = new C4String(Scope?.Name)) {
+                var collectionSpec = new C4CollectionSpec() {
+                    name = collName_.AsFLSlice(),
+                    scope = scopeName_.AsFLSlice()
+                };
+
+                ThreadSafety.DoLocked(() => {
+                    _c4coll = Native.c4db_getCollection(c4Db, collectionSpec);
```
borrrden, line 242: I have a feeling it would be a good idea to retain and release this.

src/Couchbase.Lite.Shared/API/Database/Scope.cs
```diff
@@ -1,40 +1,184 @@
...
+        #endregion
+
+        #region Internal Methods
+
+        internal bool Add(ICollection collection)
         {
-            throw new NotImplementedException();
+            var res = (collection as Collection).CreateCollection();
+            if(res)
+                (Collections as IList<ICollection>).Add(collection);
```
borrrden, line 134: This is a bit clumsy looking, it might be stylistically better to have an explicit backing variable for the property that the class can use.  Or at least some AsInternal method to avoid doing this casting all over the place.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -386,84 +404,233 @@ internal Database(C4Database* c4db)
...
+        /// </remarks>
+        /// <param name="name">The name of the scope</param>
+        /// <returns>scope object with the given scope name</returns>
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        /// <exception cref = "CouchbaseLiteException" > Thrown with <see cref="C4ErrorCode.NotFound"/>
+        /// if <see cref="Database"/> is closed</exception>
+        public Scope GetScope(string name)
         {
-            throw new NotImplementedException();
+            return _scopes.Values.SingleOrDefault(x => x.Name == name);
```
borrrden, line 510: This is going to return bad information if a scope is added via another Database instance.  I think if it is not in the cache, there needs to be a check against LiteCore.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -386,84 +404,233 @@ internal Database(C4Database* c4db)
...
+        /// <exception cref = "CouchbaseLiteException" > Thrown with <see cref="C4ErrorCode.NotFound"/>
+        /// if <see cref="Database"/> is closed</exception>
         public IReadOnlyList<ICollection> GetCollections(string scope = _defaultScopeName)
         {
-            throw new NotImplementedException();
+            ThreadSafety.DoLocked(() =>
+            {
+                CheckOpen();
+                _selectedScope = null;
+                var hasScope = _scopes.TryGetValue(scope, out _selectedScope);
```
borrrden, line 527: Same problem, if a collection is added externally it won't show up here.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -156,12 +156,21 @@ public sealed unsafe partial class Database : ICollection, IQueryFactory, IDispo
...
-        private C4Database* _c4db;
         private bool _isClosing;
         private ManualResetEventSlim _closeCondition = new ManualResetEventSlim(true);
 
+        //Pre 3.1 Database's Collection
+        private ICollection _defaultCollection = null;
+
+        //3.1+ Database
+        private ICollection _selectedCollection = null;
+        private Scope _selectedScope = null;
```
borrrden, line 171: There is no concept of "selected" scopes and collections in a database, why are these two here?

src/Couchbase.Lite.Shared/API/Database/Scope.cs
```diff
@@ -1,40 +1,184 @@
...
+        /// </summary>
+        /// <returns>All collections in this scope object</returns>
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        public IReadOnlyList<ICollection> GetCollections()
         {
-            throw new NotImplementedException();
+            ThreadSafety.DoLocked(() =>
+            {
+                if (c4Db == null) {
+                    (Collections as IList<ICollection>).Clear();
```
borrrden, line 106: These are never getting disposed

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -87,78 +147,226 @@ public void RemoveChangeListener(ListenerToken token)
...
+        /// <summary>
+        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+        /// </summary>
         public void Dispose()
         {
-            throw new NotImplementedException();
+            ThreadSafety.DoLocked(() =>
+            {
+                _disposalWatchdog.Dispose();
+                Native.c4coll_release(_c4coll);
```
borrrden, line 367: As it stands this is an unbalanced release.  The value from getCollection is *unretained*

borrrden (changes requested):
Overall the theme here is that this implementation cannot react to scopes / collections added via other Database instances on the same file.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,108 @@ public void DeleteIndex([NotNull] string name)
...
+        /// </summary>
+        internal bool IsCollectionValid()
+        {
+            bool isValid = false;
+            ThreadSafety.DoLocked(() =>
+            {
+                isValid = Native.c4coll_isValid(_c4coll);
+                if (!isValid) {
+                    isValid = Native.c4coll_isValid(_c4coll);
+                }
```
borrrden, line 359: This, if intentional, surely deserves a comment for why this is happening.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,108 @@ public void DeleteIndex([NotNull] string name)
...
+                c4db = Native.c4coll_getDatabase(c4coll);
+            });
+
+            return c4db;
+        }
+
+        private void ReleaseCollection()
+        {
+            Native.c4coll_release(_c4coll);
+            _c4coll = null;
```
borrrden, line 386: Depending on where this is called this could open a race.  Prefer safe swapping the member first, and then releasing it.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -379,89 +388,215 @@ internal Database(C4Database* c4db)
...
+        /// The default scope is an exception as it will always be listed even it doesn't contain any collection.    
         /// </summary>
-        /// <returns></returns>
+        /// <returns>
+        /// scope names of all existing scopes, in the order in which they were created.
+        /// </returns>
+        /// <exception cref="CouchbaseException">Thrown if an error condition is returned from LiteCore</exception>
+        /// <exception cref = "CouchbaseLiteException" > Thrown with <see cref="C4ErrorCode.NotFound"/>
+        /// if <see cref="Database"/> is closed</exception>
         public IReadOnlyList<Scope> GetScopes()
```
borrrden, line 456: What is the difference between this and the Scopes property?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1333,6 +1467,17 @@ internal void CheckOpenLocked()
             });
         }
 
+        internal bool HasScopeNoneCache(string scope)
```
borrrden, line 1470: Awkward naming, I had to look at the implementation to see what it does.

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Let me remove the Scope property since it's not defined in the API doc and it's not that useful to keep it around as internal either.

borrrden: Something similar going on in the scope class (Collections / GetCollections)

Sandychuang8: That Collections property under Scope I rather keep because I can get the cached collection list internally much faster than using GetCollections method.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: renamed.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Done.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Update the logic.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,109 @@ public void DeleteIndex([NotNull] string name)
...
+                c4db = Native.c4coll_getDatabase(c4coll);
+            });
+
+            return c4db;
+        }
+
+        private unsafe void ReleaseCollection()
+        {
+            IntPtr temp = IntPtr.Zero;
+            Interlocked.Exchange(ref temp, _c4coll);
```
borrrden, line 385: This doesn't actually help anything, as I mentioned, so either _c4coll needs to be an IntPtr, or just live with a simple nonatomic swap and release.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
borrrden: It seems you are checking c4coll_isValid before checking for null, this should be reversed.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -51,17 +87,33 @@ public sealed class Collection : IIndexable, IChangeObservable<CollectionChanged
...
         /// Case sensitive.
         /// </remarks>
-        public string Name => throw new NotImplementedException();
+        public string Name { get; private set; } = DefaultCollectionName;
 
         /// <summary>
         /// Gets the Scope of the Collection belongs to
         /// </summary>
-        public Scope Scope => throw new NotImplementedException();
+        public Scope Scope { get; private set; }
```
borrrden, line 95: Is this set anywhere other than in the constructor?  If not, remove private set.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -51,17 +87,33 @@ public sealed class Collection : IIndexable, IChangeObservable<CollectionChanged
         /// Cannot start with _ or %.
         /// Case sensitive.
         /// </remarks>
-        public string Name => throw new NotImplementedException();
+        public string Name { get; private set; } = DefaultCollectionName;
```
borrrden, line 90: Remove private set if this is not set anywhere other than the constructor.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -379,89 +386,217 @@ internal Database(C4Database* c4db)
...
         {
-            throw new NotImplementedException();
+            ThreadSafety.DoLocked(() =>
+            {
+                CheckOpen();
+                if (_defaultScope == null) {
+                    _defaultScope = new Scope(this);
+                    if(!_scopes.ContainsKey(_defaultScopeName))
+                        _scopes.TryAdd(_defaultScopeName, _defaultScope);
+        }
```
borrrden, line 407: Weird indentation

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,109 @@ public void DeleteIndex([NotNull] string name)
...
+
+        #region Internal Methods
+
+        internal void GetSpec()
+        {
+            ThreadSafety.DoLocked(() =>
+            {
+                var spec = Native.c4coll_getSpec(c4coll);
+                Name = spec.name.CreateString();
+                Scope.Name = spec.scope.CreateString();
```
borrrden, line 341: Why are these two being set here?  Are they expected to change since the constructor?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -379,89 +386,217 @@ internal Database(C4Database* c4db)
...
+        /// <exception cref = "CouchbaseLiteException" > Thrown with <see cref="C4ErrorCode.NotFound"/>
+        /// if <see cref="Database"/> is closed</exception>
         public IReadOnlyList<Collection> GetCollections(string scope = _defaultScopeName)
         {
-            throw new NotImplementedException();
+            Scope s = null;
+            ThreadSafety.DoLocked(() =>
+            {
+                CheckOpen();
+                s = scope == _defaultScopeName ? GetDefaultScope() : GetScope(scope);
```
borrrden, line 503: Small nit: GetScope also has the default scope, so why not just always use GetScope?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -379,89 +386,217 @@ internal Database(C4Database* c4db)
...
+                if (s != null) {
+                    s.GetCollections();
+                }
+            });
+
+            return s?.Collections;
```
borrrden, line 509: Won't this fail if Scope already exists, but collections are added to it later?

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1333,6 +1467,17 @@ internal void CheckOpenLocked()
             });
         }
 
+        internal bool HasScopeFromLiteCore(string scope)
+        {
+            bool hasScope = ThreadSafety.DoLocked(() =>
```
borrrden, line 1472: You could just return here instead of assigning to a separate variable

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Logic of this method is updated already. Not sure why "Outdated" is not showing.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,109 @@ public void DeleteIndex([NotNull] string name)
...
+        internal void CheckCollectionValid()
+        {
+            ThreadSafety.DoLocked(() =>
+            {
+                if (c4Db == null) {
+                    throw new InvalidOperationException(CouchbaseLiteErrorMessage.DBClosed);
+                }
+
+                var isValid = Native.c4coll_isValid((C4Collection*)_c4coll);
+                if (!isValid || _c4coll == IntPtr.Zero) {
```
Sandychuang8, line 357: I will switch the check order here if this is what you mean?

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Will remove this method. It's not used anywhere.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: GetDefaultScope() has much simpler logic, I need to check if LiteCore HasScope in GetScope, but default scope always exists in the database....

src/Couchbase.Lite.Shared/API/Database/Collection.cs
borrrden: Calling c4coll_isValid on nullptr is probably going to crash, that's the problem.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
borrrden: This could use some consolidation at this point, but it will suffice.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,102 @@ public void DeleteIndex([NotNull] string name)
...
+                c4db = Native.c4coll_getDatabase(c4coll);
+            });
+
+            return c4db;
+        }
+
+        private unsafe void ReleaseCollection()
+        {
+            Interlocked.Exchange(ref _c4coll, IntPtr.Zero);
+            Native.c4coll_release((C4Collection*)_c4coll);
```
borrrden, line 379: This is incorrect, `Interlocked.Exchange` will return the value that was *previously* in the ref variable.  As it stands now, this is a leak since you set _c4coll to 0, and then free _c4coll (which is zero...).  The logic should look like this:

```
var old = Interlocked.Exchange(ref _c4coll, IntPtr.Zero);
Native.c4coll_release((C4Collection *)old);
```

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -269,13 +321,102 @@ public void DeleteIndex([NotNull] string name)
...
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            if (!(obj is Collection other)) {
+                return false;
+            }
+
+            return String.Equals(Name, other.Name, StringComparison.Ordinal)
```
borrrden, line 399: What will happen if other is null?  I forget if this is a possibility here with the `is` operator.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -379,89 +386,216 @@ internal Database(C4Database* c4db)
...
+            {
+                CheckOpen();
+                var s = scope == _defaultScopeName ? GetDefaultScope() : GetScope(scope);
+                if (s != null) {
+                    var c = s.GetCollection(name);
+                    if (s.DeleteCollection(c) && s.Collections?.Count == 0) {
+                        Scope sc = null;
+                        if (_scopes.TryRemove(scope, out sc)) {
+                            sc.Dispose();
+                            sc = null;
```
borrrden, line 594: Unneeded line setting to null

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -375,8 +370,8 @@ internal void CheckCollectionValid()
 
         private unsafe void ReleaseCollection()
         {
-            Interlocked.Exchange(ref _c4coll, IntPtr.Zero);
-            Native.c4coll_release((C4Collection*)_c4coll);
+            var old = Interlocked.Exchange(ref _c4coll, IntPtr.Zero);
+            Native.c4coll_release((C4Collection*)old);
             _c4coll = IntPtr.Zero;
```
borrrden, line 375: This is redundant, it is handled by the first line.  That is what Interlocked.Exchange does, both set a value and return its old one in an atomic way.

borrrden (approved):

PR #1387, opened 2022-06-17
CBL-3228 Scopes and Collections without Document unit tests

Sandychuang8:


src/Couchbase.Lite.Tests.Shared/ScopeCollectionTest.cs
```diff
@@ -89,5 +89,277 @@ public void TestGetDefaultScopeAfterDeleteDefaultCollection()
...
+            Action badAction = (() => Db.CreateCollection(str));
+            badAction.Should().Throw<CouchbaseLiteException>($"Invalid collection name '{str}' in scope because the collection name length {str.Length} is over naming length limit.");
+        }
+
+        [Fact] // I think this test case is a dup of TestCollectionNameWithValidChars?
+        public void TestCollectionNameCaseSensitive()
+        {
+            Db.CreateCollection("COLLECTION1").Should().NotBeNull();
+            // TODO : wait for CBL-3195 fix
+            //Db.CreateCollection("collection1").Should().NotBeNull("Should be able to be created because collection name is case sensitive.");
```
borrrden, line 260: This will already return non-null, it will just be the same as COLLECTION1.  The test should ensure they are different from each other (e.g. independent documents)

borrrden (approved):

PR #1388, opened 2022-06-17
CBL-3226 Collection Implementation

Sandychuang8:
- Add Collection APIs that are deprecated in Database class

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -250,6 +250,7 @@ namespace Couchbase.Lite
         public void Purge([NotNull] string docId)
         {
             CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(docId), docId);
+            CheckCollectionValid();
             Database.InBatch(() => PurgeDocById(docId));
         }
 
@@ -321,7 +322,9 @@ namespace Couchbase.Lite
         /// doesn't exist</exception>
         public DateTimeOffset? GetDocumentExpiration(string docId)
         {
-            if (LiteCoreBridge.Check(err => Native.c4coll_getDoc(c4coll, docId, true, C4DocContentLevel.DocGetCurrentRev, err)) == null) {
+            var doc = (C4Document*)LiteCoreBridge.Check(err => Native.c4coll_getDoc(c4coll, docId, true, C4DocContentLevel.DocGetCurrentRev, err));
+            if ( doc == null) {
+                Native.c4doc_release(doc);
                 throw new CouchbaseLiteException(C4ErrorCode.NotFound);
             }
 
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -324,9 +324,10 @@ namespace Couchbase.Lite
         {
             var doc = (C4Document*)LiteCoreBridge.Check(err => Native.c4coll_getDoc(c4coll, docId, true, C4DocContentLevel.DocGetCurrentRev, err));
             if ( doc == null) {
-                Native.c4doc_release(doc);
                 throw new CouchbaseLiteException(C4ErrorCode.NotFound);
             }
+
+            Native.c4doc_release(doc);
 
             C4Error err2 = new C4Error();
             var res = (long)Native.c4coll_getDocExpiration(c4coll, docId, &err2);
--- a/src/Couchbase.Lite.Shared/API/Database/Collection.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Collection.cs
@@ -228,7 +228,6 @@ namespace Couchbase.Lite
             CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(document), document);
             ThreadSafety.DoLocked(() =>
             {
-                CheckCollectionValid();
                 VerifyCollection(document);
 
                 if (!document.Exists) {
@@ -250,7 +249,6 @@ namespace Couchbase.Lite
         public void Purge([NotNull] string docId)
         {
             CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(docId), docId);
-            CheckCollectionValid();
             Database.InBatch(() => PurgeDocById(docId));
         }
 
@@ -322,6 +320,7 @@ namespace Couchbase.Lite
         /// doesn't exist</exception>
         public DateTimeOffset? GetDocumentExpiration(string docId)
         {
+            CheckCollectionValid();
             var doc = (C4Document*)LiteCoreBridge.Check(err => Native.c4coll_getDoc(c4coll, docId, true, C4DocContentLevel.DocGetCurrentRev, err));
             if ( doc == null) {
                 throw new CouchbaseLiteException(C4ErrorCode.NotFound);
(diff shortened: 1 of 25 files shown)


src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -199,62 +204,81 @@ public bool Delete([NotNull] Document document, ConcurrencyControl concurrencyCo
...
         /// not be replicated
         /// </summary>
         /// <param name="docId">The id of the document to purge</param>
         /// <exception cref="C4ErrorCode.NotFound">Throws NOT FOUND error if the document 
         /// of the docId doesn't exist.</exception>
         public void Purge([NotNull] string docId)
         {
-            throw new NotImplementedException();
+            CBDebug.MustNotBeNull(WriteLog.To.Database, Tag, nameof(docId), docId);
+            Database.InBatch(() => PurgeDocById(docId));
```
borrrden, line 253: Missing check for valid collection?

src/Couchbase.Lite.Shared/API/Database/Collection.cs
```diff
@@ -278,12 +321,26 @@ public bool Save(MutableDocument document, Func<MutableDocument, Document, bool>
         /// doesn't exist</exception>
         public DateTimeOffset? GetDocumentExpiration(string docId)
         {
-            throw new NotImplementedException();
+            if (LiteCoreBridge.Check(err => Native.c4coll_getDoc(c4coll, docId, true, C4DocContentLevel.DocGetCurrentRev, err)) == null) {
```
borrrden, line 324: Memory leak, return value of c4coll_getDoc never released.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -753,23 +755,10 @@ public void CreateIndex([@NotNull]string name, [@NotNull]IIndex index)
...
-                    var internalOpts = indexConfig.Options;
-                    // For some reason a "using" statement here causes a compiler error
-                    try {
-                        return Native.c4db_createIndex2(c4db, name, indexConfig.ToN1QL(), indexConfig.QueryLanguage, indexConfig.IndexType, &internalOpts, err);
-                    } finally  {
-                        internalOpts.Dispose();
-                    }
-                });
-            });
+            DefaultCollection.CheckCollectionValid();
```
borrrden, line 760: This stuff all happens again anyway inside of CreateIndex, etc right?  If so, this is redundant and sort of an antipattern.  There should be no need to call CheckCollectionValid outside of the collection class.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1396,12 +1317,12 @@ internal void ResolveConflict([@NotNull]string docID, [@CanBeNull]IConflictResol
                     {
                         // Do this in a batch so that there are no changes to the document between
                         // localDoc read and remoteDoc read
-                        localDoc = new Document(this, docID);
+                        localDoc = new Document(_defaultCollection, docID);
```
borrrden, line 1320: What happens if `_defaultCollection` is null?

src/Couchbase.Lite.Shared/Query/QueryResultContext.cs
```diff
@@ -31,12 +31,18 @@ internal sealed unsafe class QueryResultContext : DocContext
...
+            : base(coll, null)
         {
             _enumerator = enumerator;
         }
 
+        //public QueryResultContext(Database db, C4QueryEnumerator* enumerator)
+        //    : base(db, null)
+        //{
+        //    _enumerator = enumerator;
+        //}
```
borrrden, line 44: Unneeded commented code

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: This is a good question. We were allowing null database before, I wonder if we are also allowing null collection as the parameter? If not, how should we handle null default collection case? Just return false in readSuccess?

Sandychuang8: Or treat doc isn't existed?

Sandychuang8: I changed to use DefaultCollection which will fetch the default collection from LiteCore.

src/Couchbase.Lite.Shared/Query/QueryResultContext.cs
Sandychuang8: removed.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Removed the checks from Database

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: released c4doc if it's found in collection.

src/Couchbase.Lite.Shared/API/Database/Collection.cs
Sandychuang8: Added valid collection check

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1396,12 +1306,12 @@ internal void ResolveConflict([@NotNull]string docID, [@CanBeNull]IConflictResol
                     {
                         // Do this in a batch so that there are no changes to the document between
                         // localDoc read and remoteDoc read
-                        localDoc = new Document(this, docID);
+                        localDoc = new Document(DefaultCollection, docID);
```
borrrden, line 1309: Still, if the default collection is deleted DefaultCollection will be returning null.  I think this should probably be a separate error condition for the legacy calls.  None of them are going to work if the default collection is deleted.  I think there should also be a test to ensure that it throws an exception instead of possibly crashing.

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Will add the test case with scenario when the default collection is deleted during conflict resolver. The expected result should throw an exception instead of possibly crashing. I added the note in this ticket: https://issues.couchbase.com/browse/CBL-3080

Sandychuang8: This ResolveConflict in Database may simply call Collection ResolveConflict after the collection replication/collection configuration are implemented I think....

borrrden: Oh I misread "deleted during conflict resolution."  This definitely sounds like it is deleted during the conflict resolver.  However, the database can have no default collection at any time.  Is the legacy API prepared to handle this?  I think this error should be coordinated with @pasin

pasin: 1. In the collection API, CouchbaseLiteException with NotOpen error code will be thrown. We could also add a custom error message such as "The default collection has been deleted."

2. In the legacy API, the behavior will be the same as when the database is deleted, which is throwing InconsistentStateException.

Sandychuang8: #1 has taken care of in different PR https://github.com/couchbase/couchbase-lite-net/pull/1390/files#diff-bd6880d0a8cd12b58b5f73a6389f6e6aa27c78e47bc346467e114b683ad4a796R676-R688
#2 has taken care of here in this commit https://github.com/couchbase/couchbase-lite-net/pull/1388/commits/e1c1d01b5c7a1b5c3b317f00a5653a694bdbeb11

borrrden (approved):

PR #1393, opened 2022-06-23
CBL-3307 Add Collection and document update observers and their tests

Sandychuang8:


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1097,23 +1070,10 @@ public void Save(Couchbase.Lite.Linq.IDocumentModel model)
         /// <exception cref="InvalidOperationException">Thrown if this method is called after the database is closed</exception>
         [Obsolete("AddChangeListener is deprecated, please use GetDefaultCollection().AddChangeListener")]
         public ListenerToken AddChangeListener([@CanBeNull] TaskScheduler scheduler,
-            [@NotNull] EventHandler<DatabaseChangedEventArgs> handler)
+            [@NotNull] EventHandler<CollectionChangedEventArgs> handler)
```
Sandychuang8, line 1073: The Database Change Listener will no longer just listen for a DatabaseChange event starts from CBL v3.1 since it should listener for a default collection change event. DatabaseChangedEventArgs will not work with the existing setup anymore. Thoughts?

borrrden: Wrap it so that it still returns databasechangedeventargs.  Document that it will only function for the default collection.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1099,27 +1072,19 @@ public void Save(Couchbase.Lite.Linq.IDocumentModel model)
         public ListenerToken AddChangeListener([@CanBeNull] TaskScheduler scheduler,
```
borrrden, line 1072: Needs `ObsoleteAttribute`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1148,24 +1113,8 @@ public void Save(Couchbase.Lite.Linq.IDocumentModel model)
         public ListenerToken AddDocumentChangeListener([@NotNull] string id, [@CanBeNull] TaskScheduler scheduler,
```
borrrden, line 1113: Also needs `ObsoleteAttribute`

borrrden (approved):
Just the two attributes needed.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: ? The Obsolete Attribute is right above.

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Same.. Obsolete Attribute is right above.

src/Couchbase.Lite.Shared/API/Database/Database.cs
borrrden: Oh the diff left out just a few lines there so I didn't notice it, sorry.

PR #1395, opened 2022-06-28
CBL-3332 Remove post changes after finishing InBatch call & Update  Scopes Collections APIs

Sandychuang8:
CBL-3332 Remove post changes after finishing InBatch call
Update/add missing Scopes Collections APIs comments and annotations

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/DatabaseChangedEventArgs.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/DatabaseChangedEventArgs.cs
@@ -51,6 +51,7 @@ namespace Couchbase.Lite
         /// <summary>
         /// [DEPRECATED]
         /// </summary>
+        [Obsolete("DatabaseChangedEventArgs is deprecated, please use CollectionChangedEventArgs")]
         internal DatabaseChangedEventArgs([NotNull] Database database, [NotNull][ItemNotNull] IReadOnlyList<string> documentIDs)
         {
             Debug.Assert(database != null);
--- a/src/Couchbase.Lite.Shared/API/Database/DatabaseChangedEventArgs.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/DatabaseChangedEventArgs.cs
@@ -25,8 +25,9 @@ using JetBrains.Annotations;
 namespace Couchbase.Lite
 {
     /// <summary>
-    /// The parameters of a database changed event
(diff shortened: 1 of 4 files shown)


src/Couchbase.Lite.Shared/API/Database/DocumentChangedEventArgs.cs
```diff
@@ -37,7 +37,7 @@ public sealed class DocumentChangedEventArgs : EventArgs
         public string DocumentID { get; }
 
         /// <summary>
-        /// The source of the document that changed
+        /// [DEPRECATED] The source of the document that changed
         /// </summary>
         [NotNull]
         public Database Database { get; }
```
borrrden, line 43: This time I am pretty sure that Obsolete is missing

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/DocumentChangedEventArgs.cs
Sandychuang8: oh! hmmm, then there are a bunch of new properties need Obsolete attributes? :s

borrrden: Possibly.  Adding deprecated to the comments is not nearly as noticeable as adding the Obsolete attribute so the attribute is the standard to use when deprecating.

Sandychuang8: Just checked and found out most of the newly deprecated properties are already contain the obsolete attributes. I added the missing ones on couple newly deprecated properties.

src/Couchbase.Lite.Shared/API/Database/DatabaseChangedEventArgs.cs
```diff
@@ -51,6 +51,7 @@ public class DatabaseChangedEventArgs : EventArgs
         /// <summary>
         /// [DEPRECATED]
         /// </summary>
+        [Obsolete("DatabaseChangedEventArgs is deprecated, please use CollectionChangedEventArgs")]
```
borrrden, line 54: It won't hurt anything, but this attribute is not as useful for internal properties, since I assume as a developer you know already

borrrden (approved):

PR #1397, opened 2022-07-01
CBL-3334 Get Collection update & Enable the tests that were disabled awaiting for the LiteCore fixes that now are in

Sandychuang8:
- Optimize the get collection logic to use cache and collection is valid LiteCore check.
- Return get collection null instead of throw in case of collection is not valid base on API spec.
- Remove has collection LiteCore check since it's replaced with collection is valid which is less expensive.
- Enable the tests that were disabled awaiting for the LiteCore fixes that now are in
- Revert the source change to take database as parameter instead of collection for query related items.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -272,7 +272,7 @@ namespace Couchbase.Lite
         {
             get {
                 GetDefaultCollection();
-                if (_defaultCollectionIsDeleted || _defaultCollection.IsClosed == true)
+                if (_defaultCollectionIsDeleted || ThreadSafety.DoLocked(() => _defaultCollection?.IsClosed) == true)
                     throw new InvalidOperationException($"Default Collection is deleted from the database {ToString()}.");
 
                 return _defaultCollection;
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -268,7 +268,18 @@ namespace Couchbase.Lite
             }
         }
 
-        internal Collection DefaultCollection => _defaultCollection == null ? GetDefaultCollection() : _defaultCollection;
+        internal Collection DefaultCollection
+        {
+            get {
+                if(_defaultCollection == null && !_defaultCollectionIsDeleted)
+                    _defaultCollection = GetDefaultCollection();
+
+                if (_defaultCollectionIsDeleted || _defaultCollection.IsClosed == true)
+                    throw new InvalidOperationException($"Default Collection is deleted from the database {ToString()}.");
+
+                return _defaultCollection;
+            }
+        }
 
         private bool IsShell { get; } //this object is borrowing the C4Database from somewhere else, so don't free C4Database at the end if isshell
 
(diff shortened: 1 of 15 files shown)


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -414,10 +414,9 @@ public Collection GetDefaultCollection()
                         WriteLog.To.Database.W(Tag, $"The none-recoverable default collection is now deleted from database {ToString()}.");
                     }
                 }
+                
+                return _defaultCollection == null || !_defaultCollection.IsValid ? null : _defaultCollection;
```
borrrden, line 418: Simplified version is `return _defaultCollection?.IsValid == true ? _defaultCollection : null`

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -496,17 +495,17 @@ public IReadOnlyList<Collection> GetCollections([@CanBeNull] string scope = _def
...
                 var s = scope == _defaultScopeName ? GetDefaultScope() : GetScope(scope);
                 if (s != null) {
-                    c = s.GetCollection(name);
+                    coll = s.GetCollection(name);
                 }
-            });
 
-            c?.CheckCollectionValid();
-            return c;
+                return coll == null || !coll.IsValid ? null : coll;
```
borrrden, line 507: Same remark for simplification.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1358,11 +1354,14 @@ private void CheckOpen()
...
 
-        private void CheckExistenceOfDefaultCollection()
+        private void CheckDefaultCollectionOpen()
         {
-            if (DefaultCollection == null) {
-                throw new InvalidOperationException($"Default Collection is deleted from the database { ToString() }.");
-            }
+            ThreadSafety.DoLocked(() =>
+            {
+                if (DefaultCollection.IsClosed) {
```
borrrden, line 1361: Not good enough.  DefaultCollection can be null and now you've thrown a null reference exception.  Why not just have your internal property (`DefaultCollection`) throw an exception if it is null or closed, rather than calling CheckDefaultCollectionOpen everywhere.

borrrden (changes requested):

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -268,7 +268,10 @@ internal bool IsClosedLocked
             }
         }
 
-        internal Collection DefaultCollection => _defaultCollection == null ? GetDefaultCollection() : _defaultCollection;
+        internal Collection DefaultCollection => _defaultCollectionIsDeleted ||
```
borrrden, line 271: There is still a risk here that I don't like.  If you call this before you ever call `GetDefaultCollection()` then this is still going to return null because that is what `GetDefaultCollection()` returns.

My suggestion is to look at the result of `GetDefaultCollection()` and if it is null then throw the exception.  

```
if(_defaultCollection == null) {
    GetDefaultCollection();
}

if(_defaultCollection == null || _defaultCollection.IsClosed) {
    throw new InvalidOperationException($"Default Collection is deleted from the database {ToString()}.")
}

return _defaultCollection;
```

borrrden (changes requested):

borrrden (approved):

PR #1407, opened 2022-07-28
CBL-3491 Named Collections Replication update with latest LiteCore (build 190+)

Sandychuang8:


Sandychuang8:
LiteCore build now is at 198. This PR is no longer up to date. Will create a new PR when we have LiteCore build number greater than 198.

PR #1418, opened 2022-08-23
CBL-3572 Add more detailed comments on enhanced certificate pinning API

Sandychuang8:


borrrden (approved):
You don't need a review for something so tiny like this

PR #1428, opened 2022-09-09
CBL-3667 Properly keep the C4DatabaseConfiguration2's parent directory (c4string) live until c4db is copied or opened

Sandychuang8:
CBL-3667 Properly keep the C4DatabaseConfiguration2's parent directory (c4string) live until c4db is copied or opened


src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -618,8 +617,10 @@ public static void Copy([@NotNull]string path, [@NotNull]string name, [@CanBeNul
                     }
                 }
                 #endif
-
-                return Native.c4db_copyNamed(path, name, &nativeConfig, err);
+                using (var parentDirectory = new C4String(config?.Directory)) {
```
borrrden, line 620: new C# is cleaner with the new using syntax (`using var parentDirectory = new C4String(config?.Directory)`)

borrrden (approved):

src/Couchbase.Lite.Shared/API/Database/Database.cs
Sandychuang8: Language version has to be 8+

PR #1431, opened 2022-09-26
CBL-3652 pac file support in android system proxy auto-config

Sandychuang8:


src/Couchbase.Lite.Shared/Support/Android/XamarinAndroidProxy.cs
```diff
@@ -32,22 +35,42 @@ internal sealed class XamarinAndroidProxy : IProxy
...
+
+            if (url == null)
+                return string.Empty;
+
+            // UriBuilder takes care of encoding everything properly
+            var bldr = new UriBuilder(url);
+            if (url.IsDefaultPort)
+                bldr.Port = -1; // Avoids adding :80 or :443 to the host name in the result
+
+            // bldr.Uri.ToString () would ruin the good job UriBuilder did
```
borrrden, line 71: lol

src/Couchbase.Lite.Shared/Support/Android/XamarinAndroidProxy.cs
```diff
@@ -32,22 +35,42 @@ internal sealed class XamarinAndroidProxy : IProxy
 
         public Task<WebProxy> CreateProxyAsync(Uri destination)
         {
-            WebProxy webproxy = null;
-            // if a proxy is enabled set it up here
-            string host = JavaSystem.GetProperty("http.proxyHost")?.TrimEnd('/');
-            string port = JavaSystem.GetProperty("http.proxyPort");
```
borrrden, line 38: Can this be kept in as a fallback?

Sandychuang8: Where is the placement you have this fallback in mind? I can only think of returning no proxy faster if host returns null...

borrrden: I'm not familiar with the way that these things get set but I'm wondering if there is a way that these vars are set without setting the system proxy so if the system proxy is null then checking these might provide backup info

borrrden: In otherwords, do what you did in the first commit. If it gives no info, check the env vars just in case.

Sandychuang8: With Android 9 API 28, I see JavaSystem.GetProperty("http.proxyHost") returns null if no system proxy nor auto-config is manually set. 
JavaSystem.GetProperty("http.proxyHost") will be set to "localhost" by the system when auto-config is manually set with a PAC file. proxyPort value will be -1 most of the time.

borrrden: Hmmmmm then maybe it's ok to just leave them off and go with only the ProxySelector.  They are based on Java runtime semantics, but Android has its own semantics anyway.

borrrden (changes requested):
I don't think I am conveying what I want to say....just remove the http.proxyHost stuff at this point, it will be simpler.

PR #1433, opened 2022-09-26
CBL-3683 nuspec files and csprojs update/clean up

Sandychuang8:


borrrden (changes requested):
LiteCore needs to be bumped before the change to 71 can happen

src/Couchbase.Lite.Support.NetDesktop/Activate.cs
```diff
@@ -75,7 +75,8 @@ public static void LoadLiteCore()
             }
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
-				if(IntPtr.Size == 4) {
+                var arch = RuntimeInformation.OSArchitecture;
+                if (arch == Architecture.X86 || arch == Architecture.Arm) {
```
Sandychuang8, line 79: Want to check if this is the correct way to get the arch from RuntimeInformation.OSArchitecture.

borrrden: I think this should be PlatformArchitecture, not OSArchitecture.

Sandychuang8: There are only ProcessArchitecture and OSArchitecture.
ProcessArchitecture - The process architecture of the currently running app.
OSArchitecture - The platform architecture on which the current app is running. 
Ref: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osarchitecture?view=net-6.0

borrrden: Sorry ProcessArchitecture is what I mean.  From the docs it seems like ProcessArchitecture is what the current runtime is (i.e. it could be x86 runtime on x64 hardware) and OSArchitecture is the architecture of the OS (which would report x64 in the previous example)

borrrden: But actually I don't think there is anything wrong with the previous check.  On 32-bit systems, IntPtr size will be 4.

borrrden (changes requested):
Just the one architecture check change, and the rest is good to go.

src/Couchbase.Lite.Support.NetDesktop/Activate.cs
Sandychuang8: I see. Since now we have x64 and arm64 process architectures, IntPtr size may not tell us additional info besides we know it's 64 bit?

borrrden: Technically I suppose it could be running on some unknown 64-bit system but I haven't yet heard of the .NET runtime on such a system so I put it in the category of "deal with it if it arises"

PR #1439, opened 2022-10-15
Replace Xamarin.Legacy.Sdk/0.2.0-alpha2 with MSBuild.Sdk.Extras in an…

Sandychuang8:
…droid support project


Sandychuang8:
Already update what is in this PR in the other PR.

PR #1440, opened 2022-10-19
CBL-3772 Add Configurations' Default Values

Sandychuang8:
PR in EE: https://github.com/couchbaselabs/couchbase-lite-net-ee/pull/80


borrrden (changes requested):
These are supposed to be public constants so that the users know what the value is.  They also need to have doc comments so that they show up in the documentation.  Since they will be public they should also be static readonly instead of const.

src/Couchbase.Lite.Shared/API/Info/Constants.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// The default is to log in a binary encoded format that is more CPU and I/O friendly and enabling plaintext is 
+        /// not recommended in production.
+        /// </summary>
+        public static readonly bool DefaultLogFileUsePlainText = false;
+
+        /// <summary>
+        /// Default MaxSize <see cref="LogFileConfiguration.MaxSize" /> property value in LogFileConfiguration.
+        /// The default max size of the log files in bytes is 52428.
+        /// </summary>
+        public static readonly long DefaultLogFileMaxSize = 52428;
```
borrrden, line 130: This value is incorrect, should be 512 * 1024 (524288)

borrrden (changes requested):

borrrden (approved):

PR #1450, opened 2022-11-22
CBL-3897 Update Maui test project to use nuget packages for Internal test job

Sandychuang8:
Remove version from net6.0 android, ios, and maccatalyst and specify version in SupportedOSPlatformVersion attribute in the csproj(s).


borrrden (approved):
A bit concerned that the CE validation is simply completely gone now.  I hope it comes back.

PR #1471, opened 2023-01-26
Update LiteCore to 3.1.0-325 and update struct for API change

borrrden:


src/LiteCore/src/LiteCore.Shared/Interop/C4ListenerTypes_defs.cs
```diff
@@ -71,6 +71,8 @@ internal unsafe struct C4ListenerConfig
         public FLSlice directory;
         private byte _allowCreateDBs;
         private byte _allowDeleteDBs;
```
bmeike, line 73: Yeah.... you're gonna need that.... ;-P

bmeike (approved):
Looks right to me

PR #1474, opened 2023-01-31
CBL-4142: Add option to allow save of parent domain cookies

borrrden:
For example, foo.bar.com could set a cookie for all of bar.com, which is normally disallowed for security reasons.


pasin (approved):

bmeike (approved):
Looks familiar and looks good.  Did you use your Defaults somewhere?

borrrden:
Not for Lithium, but I will do it for 3.1.

PR #1478, opened 2023-02-02
Update to require Android 10+ target and check for XE init

borrrden:


src/Couchbase.Lite.Tests.Android/MainActivity.cs
```diff
@@ -22,6 +22,7 @@ public class MainActivity : RunnerActivity
         protected override void OnCreate(Bundle bundle)
         {
             ActivityContext = ApplicationContext;
+            Xamarin.Essentials.Platform.Init(this, bundle);
```
pasin, line 25: Maybe in the future, we could add `bundle` as a parameter of Droid.Activate(). I'm not sure about `this` and if it's java-android, I guessed that it's an android's context (I don't know if ApplicationContext can be used instead or not).

pasin (approved):

src/Couchbase.Lite.Tests.Android/MainActivity.cs
borrrden: Hopefully in the future we won't need this library at all.  .NET 6 doesn't seem to have the same issue as Xamarin Android and the original implementation can remain.

src/Couchbase.Lite.Shared/Support/Android/XEReachability.cs
```diff
@@ -52,11 +53,20 @@ private void OnConnectivityChanged(object sender, ConnectivityChangedEventArgs e
 
         public void Start()
         {
+            if(!Droid.HasReachability) {
+                WriteLog.To.Sync.W("CouchbaseLite", "Xamarin essentials has not been initialized, disabling reachability (be sure to call Xamarin.Essentials.Platform.Init");
```
bmeike, line 57: I think you should capitalize "essentials".  Better yet, say "Xamarin.Essentials.Platform"

src/Couchbase.Lite.Tests.Android/MainActivity.cs
bmeike: So, just FWIW, this line is pretty confusing: `ActivityContext = ApplicationContext;`.  When `OnCreate` is running, the `ActivityContext` is `this`.  That thing is actually the `ApplicationContext`, which is quite, quite different beast.  I'm not clear on why you would need the application context, if you have the activity context...
What you are passing to `Xamarin.Essentials.Platform.Init` is the activity context.  It would be dead simple to see if the `ApplicationContext` will work, of course: it is available right there.  I'd bet that it *will* work and, furthermore, I'd bet that if you look at the code for `Xamarin.Essentials.Platform.Init` that about first thing that it does is to say `appContext = context.ApplicationContext`

bmeike (commented):
LGTM

PR #1480, opened 2023-02-10
CBL-4123: Add AcceptParentDomainCookies to ReplicatorConfiguration

borrrden:
Bump LiteCore to a higher version that that API changes in

### DIFF
--- a/src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
+++ b/src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
@@ -53,6 +53,7 @@ namespace Couchbase.Lite.Sync
         private const string MaxRetryIntervalKey = "maxRetryInterval";
         private const string EnableAutoPurgeKey = "autoPurge";
         private const string HeartbeatIntervalKey = "heartbeat"; //Interval in secs to send a keepalive ping
+        private const string AcceptParentDomainCookiesKey = "acceptParentDomainCookies";
 
         // HTTP options:
         private const string HeadersKey = "headers";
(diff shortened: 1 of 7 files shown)


borrrden:
Sorry about the PR validation, no idea why it's so trash lately.  Need to unwind it and probably redo it.

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
```diff
@@ -81,6 +82,17 @@ internal sealed class ReplicatorOptionsDictionary : OptionsDictionary, IDisposab
 
         #region Properties
```
bmeike, line 83: Not sure what this is for... but it looks like it is probably doing the right thing...

bmeike (approved):
Is the use of Defaults in the submodule?  I don't se the Property initialize anywhere.

LGTM

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
borrrden: You mean #region?  It's just for code collapsing in IDEs

borrrden:
Hmmm I think actually I blanked completely on the default!  Good catch.

src/Couchbase.Lite.Shared/Sync/ReplicatorOptionsDictionary.cs
bmeike: No.. I meant the whole ReplicatorOptionsDictionary class, distinct from the ReplicatorConfiguration class.  I get it now, though.  There is an analog in Java... it is just written in C in the JNI code

PR #1486, opened 2023-02-15
CBL-4170: Use Xamarin essentials for network reachability

borrrden:
Xamarin Android only


borrrden:
I structured this PR so that each commit is a grouped block of changes for easier comprehension

bmeike (approved):
How did the Reachability stuff work, before???

borrrden:
> How did the Reachability stuff work, before???

It was in the generic Reachability.cs file.  If no dependency is registered like this, it falls back to the generic one which has the xamarin runtime issue

PR #1488, opened 2023-02-16
Fix some packaging typos

borrrden:
LiteCore.framework -> LiteCore.xcframework
net6.0-android31.0 -> net6.0-android31
Condition for copying LiteCore.dll

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Info/Defaults.cs
+++ b/src/Couchbase.Lite.Shared/API/Info/Defaults.cs
@@ -78,10 +78,10 @@ namespace Couchbase.Lite.Info {
 		public static readonly int DefaultReplicatorMaxAttemptsSingleShot = 10;
 
 		/// <summary>
-		/// Default value for <see cref="ReplicatorConfiguration.MaxAttempts" /> (-1)
+		/// Default value for <see cref="ReplicatorConfiguration.MaxAttempts" /> (Int32.MaxValue)
 		/// When replicator is continuous, never give up unless explicitly stopped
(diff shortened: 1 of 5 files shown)


src/Couchbase.Lite.Shared/API/Info/Defaults.cs
```diff
@@ -78,10 +78,10 @@ public static class Constants {
 		public static readonly int DefaultReplicatorMaxAttemptsSingleShot = 10;
 
 		/// <summary>
-		/// Default value for <see cref="ReplicatorConfiguration.MaxAttempts" /> (-1)
+		/// Default value for <see cref="ReplicatorConfiguration.MaxAttempts" /> (Int32.MaxValue)
 		/// When replicator is continuous, never give up unless explicitly stopped
 		/// </summary>
```
bmeike, line 83: I think this could be a value from Defaults

bmeike (approved):
How did you discover these typos?
I have the vague feeling that I've seen this before.  If that is true, how did the typos slip by?

src/Couchbase.Lite.Shared/API/Info/Defaults.cs
borrrden: Not sure what you mean.  It is.

borrrden:
I bet you are remembering them from the actual project files and/or scripts that had a similar change.  The build was failing before the packaging step and in my flurry of rushing I didn't remember to test the packaging logic.  I have now done so.

PR #1489, opened 2023-02-21
Go through and fix all post build tests projects

borrrden:
In the process discovered significant issues with both .NET Framework and iOS / Mac Catalyst.  Those are all fixed now and all test projects have been confirmed to at least run (though they will need a new build for Apple platforms due to a packaging issue)


src/Couchbase.Lite.Tests.Shared/LogTest.cs
```diff
@@ -268,7 +268,7 @@ public void TestConsoleLoggingLevels()
             WriteLog.To.Database.W("TEST", "TEST WARNING");
             WriteLog.To.Database.E("TEST", "TEST ERROR");
             stringWriter.Flush();
-            stringWriter.ToString().Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries).Should()
+            stringWriter.ToString().Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries).Should()
```
bmeike, line 271: Huh....

bmeike (approved):
As usual, can't say much about the versioning stuff.  Looks fine.

src/Couchbase.Lite.Tests.Shared/LogTest.cs
borrrden: An FYI the API that takes a single string / char came post .NET 4.6.2.  This is the backward compatible version that takes a list of strings or chars.

PR #1490, opened 2023-02-22
Fix a test error, UWP csproj missing attribute, and bump LiteCore

borrrden:
LiteCore 3.1.0-341
Need to use the non hardcoded version of EADDRINUSE for Xamarin (both iOS and Android define \_\_MOBILE\_\_) UWP somehow is missing the C# latest version attribute (the default is 7.3, which doesn't support some of the testing code)


borrrden:
The PR validation failure is a result of the test project breaking, rather than a defect in the code (the script is running the wrong project, and will require an EE PR).  I am still untangling the things that happened in the past few months.

jianminzhao (approved):

PR #1491, opened 2023-03-02
Fix UWP tests crashing when built with .NET Native

borrrden:
This is tough to explain, but the lifetime management of the various interop objects was a bit sketchy and somehow, only in .NET Native, this caused push and/or pull filter functions defined by C# to become invalid (GC?  Memory move?) which meant that when LiteCore tried to call them it would get an unrecoverable native exception.  This PR creates a better hierarchy of objects that can have lifetimes tied to their parents.

There was also a few other issues found during this investigation dealing with tests.  The NonBlockingDatabaseOperation test was always throwing an exception and never actually verifying anything (the exception would kill the background thread and go unnoticed) and the "forum" test appears to not actually do anything and so it was removed.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -109,8 +109,6 @@ namespace Couchbase.Lite.Sync
         /// </summary>
         public X509Certificate2 ServerCertificate { get; private set; }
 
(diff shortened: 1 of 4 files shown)


borrrden:
Ignore WinUI test failures.  The project is unable to be built.  I put up a fix for that (in a separate repo where the PR validation projects live) and locally the tests all pass for that.  The other variants in the PR validation currently are passing.

bmeike (approved):

PR #1492, opened 2023-03-03
Fix a race in a P2PTest on UWP (.NET Native)

borrrden:
The weird part, as noted, is that just adding this retry loop causes the test to pass (without retrying) whereas a simple sleep for 500 ms does not accomplish the same goal.


callumbirks (approved):
That does sound weird, +1 though

src/Couchbase.Lite.Tests.Shared/P2PTest.cs
```diff
@@ -410,12 +410,30 @@ private void CollectionPushPullReplication(bool continuous)
...
+                    // Do a retry loop because there is some sort of race, but it most likely
+                    // lies in the Mock Server implementation.  The funny part is that this never
+                    // seems to actually loop, but rather its mere presence seems to cause the test
+                    // to consistently pass.  Also note, only seems to have this race issue with
+                    // UWP .NET Native.
+                    retry = false;
+                    try {
+                        // Check docs are replicated between collections colADb & colAOtherDb
+                        colAOtherDb.GetDocument("doc").GetString("str").Should().Be("string");
+                        colAOtherDb.GetDocument("doc1").GetString("str1").Should().Be("string1");
```
bmeike, line 425: How do these tests fail.  In Java, they fail by throwing an exception.  If they do that here, the failure will disappear in the retry loop?

borrrden: The assertion fails, which causes an exception.

PR #1496, opened 2023-03-09
Fix ConcurrencyTests hanging on .NET 6 iOS

borrrden:
The synchronization context of the runtime is different than the others which seems to cause a deadlock when running WaitAssert.RunAssertAsync without using `ConfigureAwait(false)`

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/ConcurrencyTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/ConcurrencyTest.cs
@@ -152,8 +152,8 @@ namespace Test
 
             var docIDs = CreateDocs(nDocs, "Create").Select(x => x.Id).ToList();
 
-            var t1 = Task.Factory.StartNew(() => ReadDocs(docIDs, nRounds));
-            var t2 = Task.Factory.StartNew(() => UpdateDocs(docIDs, nRounds, tag));
+            var t1 = Task.Run(() => ReadDocs(docIDs, nRounds));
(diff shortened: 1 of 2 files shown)


src/Couchbase.Lite.Tests.Shared/ConcurrencyTest.cs
```diff
@@ -468,9 +467,7 @@ private void ConcurrentRuns(uint nRuns, Action<uint> block)
                 var ignore = expectations[i].RunAssertAsync(block, i);
             }
 
-            foreach (var exp in expectations) {
-                exp.WaitForResult(TimeSpan.FromSeconds(60));
-            }
+            WaitAssert.WaitFor(TimeSpan.FromSeconds(60), expectations);
```
bmeike, line 470: Yeah.   I would say...

src/Couchbase.Lite.Tests.Shared/Util/WaitAssert.cs
```diff
@@ -31,8 +32,9 @@ public sealed class WaitAssert
 
         public static void WaitFor(TimeSpan timeout, params WaitAssert[] asserts)
         {
-            foreach (var assert in asserts) {
-                assert.WaitForResult(timeout);
+            var handles = asserts.Select(x => x._mre).ToArray();
+            if (!WaitHandle.WaitAll(handles, timeout)) {
```
bmeike, line 36: While this smells bad, I should think that it would actually work... that it would not hang.  I'd be interested to know the conditions under which it did.

bmeike (approved):
Yeah... Looks good.

PR #1497, opened 2023-03-14
CBL-4289: Don't skip waiting for pending conflict tasks

borrrden:
Previous code used to skip the conflict task check if there was no recorded error.  This seems like a meaningless check, so remove it and always check.  Also add the proper task into the conflict tasks list.  Previously the "continue with" tasks was being added, but the actual work task was being removed, which means the list would grow forever (full of completed tasks).

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -599,8 +599,13 @@ namespace Couchbase.Lite.Sync
 
                     _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, false));
                 });
+                    
+                t.ContinueWith(task =>
+                {
+                    _conflictTasks.TryRemove(task, out var dummy);
+                });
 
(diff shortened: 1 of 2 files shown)


borrrden:
iOS failed because it lost the ability to correctly sign the test harness app.  I have fixed that situation, and re-run the tests manually and they all pass.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -599,8 +599,13 @@ private void OnDocEndedWithConflict(List<ReplicatedDocument> replications)
...
                     _documentEndedUpdate.Fire(this, new DocumentReplicationEventArgs(new[] { replication }, false));
                 });
+                    
+                t.ContinueWith(task =>
+                {
+                    _conflictTasks.TryRemove(task, out var dummy);
+                });
 
-                _conflictTasks.TryAdd(t.ContinueWith(task => _conflictTasks.TryRemove(t, out var dummy)), 0);
+                _conflictTasks.TryAdd(t, 0);
```
pasin, line 608: Very good catch!

pasin (approved):

PR #1499, opened 2023-03-18
CBL-4317: Deprecate ReplicatorConfiguration.Database

borrrden:


borrrden:
No idea what the failure is about (seems like the tests got stuck) but it's not related to this PR.

pasin (approved):

bmeike (commented):
I want to say that I've found this deprecation inconvenient, when writing demo apps.

A replicator configuration *cannot* involve more than one DB.  Both the replicator and its configuration know what database it is.  It seems a bit obstructionist, to refuse to tell...

PR #1504, opened 2023-03-28
CBL-4342: Deprecate ReplicatorConfiguration.Database

borrrden:


bmeike (approved):
I think this change is a mistake.  Not this change, the design.
A replicator *MUST* know the database that it is replicating.  Refusing to tell the user (she has to get the collections, get the first one, and ask it for it's database) is just mean.

pasin:
The ReplicatorConfiguration's database property is deprecated in corresponding to the deprecating of the ReplicatorConfiguration that takes the database. So I think that action is appropriate. 

If it's important to know the database that the replicator is using, we can consider putting it at the Replicator class instead. Is there a particular use-case to get the database from the replicator to use it?

PR #1505, opened 2023-03-29
Tweak which tests get run on .NET 6 Android

borrrden:
Actual code fixes in EE repo.  The status is that dynamically generated X509 certificates will never function until .NET 8.  Static certificates not trusted by default will need a network security config entry.  Intermediate certificate pinning is also not going to work, and no fix for that seems to be scheduled yet.

Ref: https://github.com/dotnet/runtime/issues/84202 and https://github.com/dotnet/runtime/issues/45741


jianminzhao (approved):

bmeike:
Is this something we need to document?  Should we create a ticket, right now, to remove these exceptions?

bmeike (approved):

borrrden:
We will absolutely need to document the life out of this.  .NET 6 iOS and Android will need their own quirks and shortcomings section (respectively).  I'll try to find the appropriate place to put a docs comment.

As for a ticket we can file it but it will remain blocked for probably months.

PR #1509, opened 2023-04-13
CBL-4421: Properly freeze collection configuration after used in Replicator

borrrden:
The old implementation mistakenly believed that calling `Freeze()` was enough, but that method returns a frozen copy, and doesn't modify in place.


bmeike (approved):
For goodness sake....

PR #1510, opened 2023-04-13
CBL-4419: Live queries built using DataSource.Collection don't get added to Database Stoppable list

borrrden:
The heart of the problem was the methods to add and remove active stoppables in QueryBase.  In the case of DataSource.Collection, the Database property will be null and so this logic was silently skipped.


bmeike (approved):
I'm having a hard time reading some of this, but it looks ok.

What is _legacyColumn about?

borrrden:
_legacyColumn means it was made using the old `DataSource.Database` input, which needs to have the column named after the database and not its default collection.

PR #1512, opened 2023-04-13
CBL-4426: Correct an error with disposing the default collection

borrrden:
GetDefaultCollection will now return a new object if it detects that the previously stored one has been invalidated by a dispose


bmeike (approved):
Yep

pasin (approved):

PR #1515, opened 2023-04-18
CBL-4427: Correct an error with disposing the default collection

borrrden:
GetDefaultCollection will now return a new object if it detects that the previously stored one has been invalidated by a dispose

This is ported from 0b179984e03cb43be5af6efa50af8ca7745443e7 via cherry pick plus extra changes (removing a redundant variable and its usage.  The change was too large for the release branch)


velicuvlad (approved):
LGTM

pasin (approved):

PR #1516, opened 2023-04-21
CBL-4367: Migrate to nullable reference types

borrrden:
This change is gigantic, but the bulk of it is just a lot of adding question marks to type names.  Like Swift, C# has the ability to say that a variable is never meant to be null, unless marked with a question mark.  Despite the mass amount of changes, there is no direct functionality change.  It is purely for built in static analysis and the compiler will issue warnings when it cannot confirm that a non null variable is always null in all cases.  

This identified some areas where a bit more care was needed so I added some various null checks and/or asserts.  Also oddly I ran into a recursive call problem with Database.Log that was solved by not using CBDebug inside of Service.AutoRegister (Because Log calls GetRequiredInstance for IConsoleLogger, which then calls back to Log which is null because the original call is not done yet).


velicuvlad (approved):
LGTM

src/Couchbase.Lite.Shared/API/Query/IResultSet.cs
```diff
@@ -30,13 +29,13 @@ namespace Couchbase.Lite.Query
...
-    public interface IResultSet : IEnumerable<Result>
+    public interface IResultSet : IEnumerable<Result?>
     {
         /// <summary>
         /// Cross platform API entry to get all results in a list.  Same
         /// as <c>ToList()</c>
         /// </summary>
         /// <returns>A list of results from the result set</returns>
-        List<Result> AllResults();
+        List<Result?> AllResults();
```
pasin, line 39: I think the Result cannot be null

src/Couchbase.Lite.Shared/API/Query/Parameters.cs
```diff
@@ -130,8 +126,7 @@ public Parameters SetBoolean([NotNull]string name, bool value)
         /// <param name="name">The name of the key to set</param>
         /// <param name="value">The value to set</param>
         /// <returns>The parameters object for further processing</returns>
-        [NotNull]
-        public Parameters SetDate([NotNull]string name, [CanBeNull]DateTimeOffset value)
+        public Parameters SetDate(string name, DateTimeOffset value)
```
pasin, line 129: DateTimeOffset is CanBeNull.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
```diff
@@ -190,8 +184,7 @@ public IList<string> DocumentIDs
         /// <summary>
         /// Extra HTTP headers to send in all requests to the remote target
         /// </summary>
-        [NotNull]
-        public IDictionary<string, string> Headers
+        public IDictionary<string, string?> Headers
```
pasin, line 187: Headers is nullable.

src/Couchbase.Lite.Shared/Query/NullResultSet.cs
```diff
@@ -69,7 +69,7 @@ public void Reset()
 
         #region IResultSet
 
-        public List<Result> AllResults() => new List<Result>();
+        public List<Result?> AllResults() => new List<Result?>();
```
pasin, line 72: Not sure how this works but there shouldn't be null Result in the list.

pasin (changes requested):

src/Couchbase.Lite.Shared/API/Query/Parameters.cs
borrrden: DateTimeOffset cannot be null, this attribute was wrong.  It is a struct, so the compiler does not allow null.

src/Couchbase.Lite.Shared/API/Sync/ReplicatorConfiguration.cs
borrrden: I don't like allowing null collections....there is no point and just adds extra null checks.  This was notnull before according to the attribute.

src/Couchbase.Lite.Shared/API/Database/Database.cs
```diff
@@ -1111,9 +1115,9 @@ internal void RemoveActiveStoppable(IStoppable stoppable)
             });
         }
 
-        internal string GetCookies([@NotNull] Uri uri)
+        internal string? GetCookies(Uri? uri)
```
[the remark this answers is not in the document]
borrrden: There is not, but analysis of the code shows that null is handled without throwing an exception.  Changing this would be kind of breaking I guess?

src/Couchbase.Lite.Shared/API/Query/IResultSet.cs
borrrden: I think I can make this work actually

pasin (approved):

PR #1520, opened 2023-05-02
CBL-4434: Address all compiler warnings project wide

borrrden:


borrrden:
Some of the warnings were regarding missing / corrupted documentation comments which is why you see those in there as well

borrrden:
The failed test is known to be flaky, and unrelated to this PR.

pasin (approved):

PR #1521, opened 2023-05-09
CBL-4273: Fix MacProxy to account for native library path changes

borrrden:
CFNetwork changed locations in macOS 13


borrrden:
Note to reviewer:  This change was made by performing the following two steps:

1. Copy and paste the MacProxy.cs file from master
2. Change the top level `#if` back to the old defines to reflect the versions of .NET being built for

pasin (approved):

PR #1531, opened 2023-05-24
CBL-4504: Add attributes to iOS proxy unmanaged delegates

borrrden:
Without this, the AOT compiler will not realize them and they will fail at runtime


borrrden:
Will have to look at the Jenkins job yet again to see why iOS is failing, because it's certainly not from this tiny change

bmeike (approved):
If yo say so... ;-P

PR #1532, opened 2023-05-25
Update LiteCore to 3.1.1-3

borrrden:


borrrden:
I will look at the mac machine today to find out why iOS testing is not functioning correctly

bmeike:
These test failures are, really, such a bore.  Is there anything we can do to make them more reliable?

bmeike (approved):

borrrden:
Yes, but the attempts usually don't last long...

PR #1535, opened 2023-06-02
CBL-4585: Fix a regression caused by removing Microsoft.Bcl.AsyncInte…

borrrden:
…rfaces

This was a botched dependency graph in the minimum version of SimpleInjector that we required.  Bump the minimum to 5.2.1 which is the first GA version that doesn't require that dependency at all anymore.


borrrden:
PR validation is messed up by not having an updated Jenkinsfile in THIS repo....see the companion EE PR for an actual run

bmeike (approved):

PR #1536, opened 2023-06-03
Regenerate expired client.p12 cert

borrrden:
And its issuer client-ca.der since the key is nowhere to be found.  Also include a script to regenerate them again in the future.


borrrden:
All Windows tests are passing now.  Xamarin remains flaky as usual.....

borrrden:
I analyzed the three test failures.  One is a test mistake that can occur by running with a newer version, then an older version of the library without deleting the app (database too new error) and the other two I ran in isolation and they both passed multiple times.

bmeike (approved):

PR #1539, opened 2023-06-06
CBL-4587: Fix a regression caused by removing Microsoft.Bcl.AsyncInterfaces

borrrden:
This was a botched dependency graph in the minimum version of SimpleInjector that we required.  Bump the minimum to 5.2.1 which is the first GA version that doesn't require that dependency at all anymore.


borrrden:
Windows here will not pass without an EE change, see the companion EE PR for this

borrrden:
Ported manually from 15f026633f2da0f53de2b85779447bdb48839d7c (basically just changing version numbers) with some test fix stuff on top (regen certs, etc)

pasin (approved):
Assume that PR validation failure is not relevant.

borrrden:
It was an expired run.  The same PR on master was clean.

PR #1541, opened 2023-06-07
CBL-4578: Make IReachability public

borrrden:
This won't have any bearing on most users, but in rare cases they may want to provide their own reachability logic in guided support cases


pasin (commented):
If I implement a custom reachability, how do I set it to the replicator? Shall we have this feature cross platforms?

pasin (approved):

borrrden:
Addressed offline.  This is meant to be used in guided scenarios.

PR #1542, opened 2023-06-07
CBL-4577: Make IReachability public

borrrden:
This won't have any bearing on most users, but in rare cases they may want to provide their own reachability logic in guided support cases


borrrden:
Same PR passed validation on master.  These tests need some work (as in, I did something to make them work well on master but pre-master needs those changes).

pasin (approved):

PR #1546, opened 2023-06-21
CBL-4613: Upgrade .NET non-windows to .NET 7

borrrden:
Since .NET has a different Microsoft support lifecycle for these targets (.NET 6 already out of support)


borrrden:
One test is being flaky, which is better than the 5 that it used to be.  Regardless it is not related to this PR.

bmeike (approved):
my goodness, what fiddly bs.  ;-(

PR #1547, opened 2023-06-22
Try to plug a race between setting and waiting in a test

borrrden:


borrrden:
Maui iOS will depend on fixing something in EE (better) or fixing something here (acceptable).

bmeike (approved):

PR #1548, opened 2023-06-22
Update to LiteCore 3.1.1-9

borrrden:
Add some test stability stuff in as a bonus


borrrden:
.NET stuff is a mystery, but I double checked the exact same steps on another machine and all tests passed with no issue.

bmeike (approved):

PR #1550, opened 2023-06-28
CBL-4526: Enable tests that were disabled on Android / Apple

borrrden:
Also, update LiteCore so that there is no need for a separate PR review


borrrden:
Finally iOS is more reliably passing, but by coincidence a test that fails only once in a long while failed on Windows desktop (then went on to pass in basically the same environment aka WinUI).  I won't address that in this PR, and I think the goal of the ticket is accomplished and more.

jianminzhao (approved):

PR #1553, opened 2023-07-07
CBL-4652: Don't allow calls to API after object disposed

borrrden:
This will prevent situations in which the native object is recreated, and then never disposed because dispose was already called


callumbirks (approved):
With a brief assumption of how the disposal system works, LGTM

PR #1555, opened 2023-08-01
Update LiteCore to 3.2.0-83

borrrden:
Change API of c4query_run to match, and update LogTest for updated log metadata


bmeike (approved):
yep!

PR #1556, opened 2023-08-21
Finish removing the defunct LiteCore tests

borrrden:


velicuvlad:
are these not used anymore or are some sort of tombstones from old days that moved to Core?

velicuvlad (approved):
lgtm

borrrden:
These tests were an idea that I had to make sure the bindings were functioning correctly by mirroring the LiteCore tests but they have the downside of

1. Being hard to maintain as the LiteCore API changes
2. Artificially keeps LiteCore API alive in the bindings when they are only used by the tests

There has never once been an issue with bindings and they are very thin so I made the decision that they don't provide enough value to keep around.

PR #1565, opened 2023-09-21
CBL-4891: Backport a hanging test fix and log test starts

borrrden:
The multiple replicators test had some flaky logic in it that sometimes caused that particular test to hang and endless print PING PONG every 5 minutes.  Backported https://github.com/couchbase/couchbase-lite-net/commit/8efacff992e134d4039c4082259e33c10bd4148d to thie branch.  The tests still hung after that, however, so I added in some logging for when a test begins.  This alone caused the hang to stop (or maybe it just didn't feel like hanging today) but it's a good change to keep just in case the hangs come back.


jianminzhao (approved):

bmeike (commented):
✅

PR #1582, opened 2024-01-25
CBL-5272: Protect c4queryobs_setEnabled(true) with db-exclusive lock

borrrden:


src/Couchbase.Lite.Shared/Query/LiveQuerier.cs
```diff
@@ -156,6 +164,17 @@ private unsafe void QueryObserverCalled(C4QueryObserver* obs, C4Query* query)
             return null;
         }
 
+        private ThreadSafety DatabaseThreadSafety()
+        {
+            switch(_queryBase.Database?.ThreadSafety) {
+                case null:
+                    WriteLog.To.Query.W(Tag, "Database thread safety unavailable, using instance thread safety...");
```
bmeike, line 171: Is this really the right thing to do?
It surely wouldn't guarantee visibility, in the Java MM.  Seems like it might not guarantee atomicity, either...

bmeike (approved):
If DatabaseThreadSafety is really the right thing, then this is good.

src/Couchbase.Lite.Shared/Query/LiveQuerier.cs
borrrden: Not sure I understand what you mean.  I'm trying to use the database exclusive lock, but I'm scared what will happen if Database is null so I fall back to the query's own lock (which isn't correct but it's not a null reference exception) which is the current behavior

borrrden: Oh and as an addendum, the live query tests all had a non null database.

bmeike: Right.... so is the inability to get the db lock actually a recoverable situation?  I can see providing a clearer exception than an NPE... but isn't continuing just whistling in the wind?

borrrden: Now I've gone and looked and convinced myself that there is never a situation in which a query will be without a database (or a collection at least) and the only reason that they are nullable is because of the query builder setting them in a fluent fashion so an exception seems more appropriate.

bmeike: Awesome.  Very cool!

bmeike (approved):

PR #1583, opened 2024-01-31
CBL-5312: Correct ValueIndexConfiguration to actually make ValueIndex

borrrden:
Add a test to assure that is the case


borrrden:
Test failures are a recurring theme at the moment, I'll make a ticket to look at that. It's always two tests on .NET iOS, not related to this PR.

bmeike (approved):
That's pretty funny.   I had several similar issues.... I think you found 2 of them

PR #1585, opened 2024-02-01
CBL-5006: Parse any number of version digits

borrrden:
Before it was strictly three, which made 3.1.3.1 appear as 1.3.1 (for example).


bmeike (approved):
heh.... I just did something similar...  only for bash

PR #1587, opened 2024-02-12
CBL-5271: Protect c4queryobs_setEnabled(true) with db-exclusive lock

borrrden:


borrrden:
Cherry picked from 22d127150552a6012a5912d2ddac9315fedd67ac

borrrden:
Same two tests failing on Jenkins MAUI iOS but not related to this (actually while doing some research I did some tests with this and they all passed for iOS)

bmeike (approved):
Relieved to find that I also do this....

PR #1588, opened 2024-02-12
CBL-5356: Fix incorrect FLDictIterator loop

borrrden:
The termination condition is when key is null, not when next returns false


borrrden:
Not sure what happened to our CI infra here.  I ran the tests locally for iOS and they all passed except for one which is not a completely valid test to begin with (TestQueryExpressions is testing 16 digits of double precision when only 15 are guaranteed).  I modified that test and it passed as well.

bmeike (approved):
Ohhhhkaaay....

PR #1594, opened 2024-02-24
CBL-5430: Make Visual Studio search more reliable

borrrden:
And also make sure we continue building this line with the .NET 6.0.x SDK


borrrden:
PR validation irrelevant here since this only applies to build server logic

bmeike (approved):

PR #1596, opened 2024-02-28
Rework the way that platform defines are handled

borrrden:
This makes it easier to switch between .NET versions


borrrden:
I'll do this again at a later time since there are fewer platforms coming soon

PR #1598, opened 2024-03-01
CBL-5447: Add in required functionality to load and run vector search…

borrrden:
… extensions

The API is EE so it it not in this repo.  A few things needed to be done in order to accomplish this:

1. The static database constructor now searches for plugins (though I may change this since a call to activate is currently required anyway, to make things a bit more understandable)
2. Remove defunct error message constant (needed to add the functionality that it claimed was illegal)
3. Add in defaults for some vector search parameters
4. Add the ability to add collections to query parameters
5. Fix up some packaging stuff in Android and WinUI
6. Add the new C4 API to the list of methods that .NET uses, and generate new bindings
7. Remove Unimplemented from the list of bugs that attach a "please file a bug report" message since we now use it to indicate lack of extensions.

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -37,6 +37,8 @@ using LiteCore.Util;
 
 using System.Collections.Concurrent;
 using System.Diagnostics;
+using System.Reflection;
+using Couchbase.Lite.Extensions;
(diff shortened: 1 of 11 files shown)


pasin (approved):
Parameters.setBlob() is also missing but that is not releated to VS.

borrrden:
@pasin That function was already present for some reason

PR #1600, opened 2024-03-06
Revert extensions changes

borrrden:


bmeike (approved):
Is this just overkill for now?

borrrden:
I decided to take a different route instead.  I didn't like the awkward placement as even I fell victim to it.  I wondered why my tests were claiming that vector search was not loaded, forgetting that I had set up a custom logger...the new way will allow more leeway before opening a database to activate the plugin.

PR #1605, opened 2024-03-12
CBL-5505: Re-enable Vector Search functionality

borrrden:
Here is it just a couple constants, since all the logic is EE.

Side: FleeceMutableArray null handling


borrrden:
Another double repo commit so this PR validation will fail

src/Couchbase.Lite.Shared/API/Document/ArrayObject.cs
```diff
@@ -130,7 +130,6 @@ internal MCollection ToMCollection()
 
         private static MValue Get(FleeceMutableArray array, int index, IThreadSafety? threadSafety = null)
         {
-            array.Context.CheckDisposed();
```
borrrden, line 133: No longer needed since the underlying Get call will verify this, making this call redundant

pasin (approved):

bmeike (approved):

PR #1607, opened 2024-03-21
CBL-4452: Migrate to nullable reference types in tests

borrrden:
Massive amount of lines changed but 95% of it is just removing `#nullable disabled` and then addressing the various `?` and `!` needed.  I tried to be mostly diligent about adding a null check to the test before adding `!` but toward the end and the 1000th time doing it for basically the same functionality I gave up.

CBL-5547
This uncovered three nullability regressions though (you can see them in the non test files).  I filed one as a ticket, but will cover all three with the same ticket.

I also added in an assembly property to not auto append the git SHA to the version, so that it remains pure as x.y.z-bdddd


bmeike (approved):
Not going to read this carefully:  The whole point, really, is getting more automatic checks.  Definitely a good thing to do.

pasin (approved):

PR #1615, opened 2024-04-19
CBL-5648: Dispose the context handle of LiveQuerier

borrrden:
Otherwise this class will live forever as a memory leak.  Also tighten up the disposal protections on this class.


bmeike (approved):
Yeah... ok.
To me, this seems like it might be more complicated than it needs to be.  In Dispose, I think it might be sufficient to free the queryObserver if it is not null and the contextHandle if it is allocated.  That's nearly all the way to getting rid of the disposed flag.

PR #1620, opened 2024-04-29
CBL-5566: Replicator ID markings in logs

borrrden:
.NET was already doing most of this, aside from an o for representing "one-shot"

Along for the ride in this commit is a nullability regression fix


bmeike (commented):
👍🏻

bmeike (approved):
I thought I did this!

PR #1621, opened 2024-05-08
CBL-5727: Throw exception instead of native crash with ColumnNames

borrrden:
Accessing ColumnNames after dispose will cause a native null deference without this fix.


borrrden:
Still waiting on a new machine to have any hope of validating beryllium commits in an automated fashion.

bmeike (approved):
Hmmm... I better check to see if I need to do this too....

PR #1622, opened 2024-05-15
CBL-5747: Don't use default value for Index null locale

borrrden:
It needs to stay null otherwise the stemming might have unexpected results if the default happens to be "en".


bmeike (approved):
Presuming that TwoLetterISOLanguageName can return null, this looks great.  I'm surprised, though, that the compiler didn't complain about the previous definition of _locale

borrrden:
`TwoLetterISOLanguageName` does NOT return null.  The previous behavior was to default to this, and let the user change to null if they like.  A previous commit I made disallowed this and when the user set null it went back to `TwoLetterISOLanguageName` instead of being `null`

PR #1625, opened 2024-05-24
CBL-5781: Create a sanity suite of tests to run on tier 2 platforms

borrrden:
The tests to exclude are only about 40, but those 40 account for about half of the test runtime (usually due to artificial thread sleeping and such).

Tier 2 platforms are largely judged by their re-use of an identical LiteCore on another platform:

Xamarin Android -> Same as .NET Android
Xamarin iOS -> Same as .NET iOS
.NET Mac Catalyst -> Same as .NET iOS
.NET WinUI -> Same as .NET 6
.NET Framework -> Same as .NET 6


borrrden:
Oh and also, I found some .NET 4.6.2 failures that either are not showing up on Jenkins, or are just silently being ignored 😩 .  Fixed the tests to eliminate that badness.

borrrden:
The scripts for PR validation have not been updated for the new machine yet, so it's going to fail.

PR #1628, opened 2024-06-18
 CBL-5749: Implement lazy index API for vector index 

borrrden:
There are no PR validations tests around this, but it was tested using several toy builds


borrrden:
You can ignore the changes in the "Interop" folder.   Just some adjustments to the auto gen.

pasin (approved):

PR #1631, opened 2024-06-20
Fix an apparently longstanding issue with core boolean methods

borrrden:
If they return false with an error code of 0, the platform interprets that (incorrectly) as true


bmeike (approved):
I have had so much trouble with this stuff.  The party line is: "It works like this except when it doesn't.  'When it doesn't' is documented in comments"  Its a PITA

PR #1632, opened 2024-06-20
Fix an apparently longstanding issue with core boolean methods

borrrden:
If they return false with an error code of 0, the platform interprets that (incorrectly) as true


borrrden:
Cherry picked without modification from bf0787e2c9a10db64a02d6583890a07e28765674

bmeike (approved):

PR #1634, opened 2024-06-24
Allow missing index error to be handled in GetIndex

borrrden:
It's not an error condition, so null should be returned.


bmeike (approved):
What does AllowError do?  That's pretty cool

borrrden:
AllowError is just what its name suggests.  If that particular error is returned, it is not considered for throwing an exception, and is instead ignored and the return value used.

PR #1635, opened 2024-06-25
CBL-5789: Fix issue where exception in save causes memory leak

borrrden:


bmeike (approved):
Yeah.  I had a couple of these....

PR #1637, opened 2024-06-28
CBL-5857: Rename Extension.Load to Extension.Enable

borrrden:
Use the new LiteCore API to return some early error information about it.

### DIFF
--- a/core_version.ini
+++ b/core_version.ini
@@ -1,6 +1,6 @@
 [version]
-build = 3.2.0-195
+build = 3.2.0-210
 
 [hashes]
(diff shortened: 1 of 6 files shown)


borrrden:
PR validation is meaningless since it needs an EE API update

borrrden:
Also note that most of this PR is moving a few functions from one file to another since they were moved in Core, and the generation script mirrors the naming of the core headers.

pasin (approved):

bmeike (approved):

PR #1641, opened 2024-07-11
Update DistanceMetric to new default

borrrden:


bmeike (approved):
Yep.  Definitely trivial

PR #1644, opened 2024-07-16
CBL-5790: Fix issue where exception in save causes memory leak

borrrden:
Not quite a cherry pick of 7184cc8272631a04d3a998ddb294a2ae2c1d3a44 , but it is a literal copy paste from the newer source into the correct places.


bmeike (approved):
Wow.... Took me a while to find the actual change!

pasin (approved):

PR #1646, opened 2024-07-24
Bump LiteCore to 3.1.9-1

borrrden:


borrrden:
Note that this is not the final core, but the current core is unusable because of orphaned commits.

bmeike (approved):
Is this worth doing?  We are supposed to have a 3.1.9-2 very very shortly...

borrrden:
Normally no, but I wanted to try out a build to see what issues happened with going back to the 3.1 line.  However it's not possible now because of the orphaned commits in SQLiteCpp.

PR #1647, opened 2024-07-30
CBL-6118: Fix path in uwp.targets

borrrden:


pasin (approved):

bmeike (approved):
This must get boring....

PR #1649, opened 2024-07-31
CBL-6073: Implement FullSync feature (and update Core to 3.2.0-223)

borrrden:
This differs moderately from the 3.1 counterpart in the following ways for the tests (the actual logic, while not cherry picked, is identical via copy/paste):

1. 3.2 and newer don't expose c4db_getConfig2 directly from the library, but the same technique can be used in the test itself to resolve the native symbol (see TestNative in the new test file)
2. Theory is used instead of local function now that UWP is gone (it used a different underlying testing framework that didn't understand Theory)

### DIFF
--- a/src/Couchbase.Lite.Tests.Shared/SQLiteOptionsTest.cs
+++ b/src/Couchbase.Lite.Tests.Shared/SQLiteOptionsTest.cs
@@ -1,0 +1,103 @@
+//
+//  SQLiteOptionsTest.cs
+//
+//  Copyright (c) 2024 Couchbase, Inc All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


src/Couchbase.Lite.Tests.Shared/SQLiteOptionsTest.cs
```diff
@@ -0,0 +1,103 @@
...
+        ///     6. Get the config FullSync property and verify that it is false
+        /// </summary>
+        [Fact]
+        public unsafe void TestSQLiteFullSyncConfig()
+        {
+            var config = new DatabaseConfiguration();
+            config.FullSync.Should().BeFalse("because the default should be false");
+
+            config.FullSync = true;
+            config.FullSync.Should().BeTrue("because C# properties should work...");
```
bmeike, line 59: lol

bmeike (approved):

PR #1654, opened 2024-08-06
Forgot that 3.1 doesn't have nullability available

borrrden:
That means that the build fails on the project when using things like `?` and `!`


borrrden:
This fixes a regression in the previous commit

bmeike (approved):

PR #1656, opened 2024-09-03
CBL-5182: API for array indexing

borrrden:


src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
```diff
@@ -28,53 +31,48 @@ namespace Couchbase.Lite.Internal.Query
...
-        internal IndexConfiguration(C4IndexType indexType, C4QueryLanguage queryLanguage)
+        internal IndexConfiguration(C4IndexType indexType, string[]? expressions)
         {
+            if(expressions != null) {
+                // Quick sanity check.  If I allow this to proceed, no error will happen but
+                // perhaps unintended behavior will.  The final string to pass will be empty
+                // and LiteCore will index the entire path element.  This is ok in some cases
+                // like ArrayIndexConfiguration, but it should be manifest with a null array
+                // and not an empty array or array of empty or array of null.
+                if(!expressions.Any(x => !String.IsNullOrEmpty(x))) {
```
bmeike, line 53: I believe we have been explicitly forbidden to make this check.
I was told to remove it.  I have been told that all of the platform APIs should behave in the same way.
You do you.

bmeike (approved):

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
borrrden: I could simplify this further but the situation I wanted to avoid was the user passing a single empty string, which would be the same as not passing any expressions at all and it's weird to have two ways to do the same thing in the first place, but the latter feels like not something a user would do on purpose but rather by error.  An error that will not cause any errors in LiteCore.

borrrden: Ok I changed the check to more accurately reflect the narrow case that I want to check.

src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
```diff
@@ -28,53 +31,48 @@ namespace Couchbase.Lite.Internal.Query
...
-        internal IndexConfiguration(C4IndexType indexType, C4QueryLanguage queryLanguage)
+        internal IndexConfiguration(C4IndexType indexType, string[]? expressions)
         {
+            if(expressions != null) {
+                // Quick sanity check.  If I allow this to proceed, no error will happen but
+                // perhaps unintended behavior will.  The final string to pass will be empty
+                // and LiteCore will index the entire path element.  This is ok in some cases
+                // like ArrayIndexConfiguration, but it should be manifest with a null array
+                // and not an empty array or array of empty or array of null.
+                if(expressions.Length == 1 && expressions[0] == "") {
```
bmeike, line 53: Yeah.... Good call.  I noticed something similar that I want to discuss this afternoon.

bmeike (approved):

PR #1658, opened 2024-09-11
Epic thread safety refactor for LiteCore

borrrden:
Get ready for a novel.  Since we have started annotating the native API with thread safety requirements, I thought it fitting to move all of the enforcement of those rules as close to the native calls as possible.  Let's start with a little background

== How IDisposable Works ==

In C#, when you have unmanaged objects that need cleanup, the accepted way of doing that is to implement `IDisposable`.  This exposes a `Dispose` method in which you can immediately free your managed and unmanaged resources.  The implication being that you would both free your unmanaged resources, and call `Dispose` on any `IDisposable` items that you happen to manage.  This integrates with the language `using` keyword, which will automatically dispose your resource regardless of what happens (return early, exception, etc).

However, sometimes people forget to call `Dispose` and so a finalizer is also implemented.  This is intended to cleanup ONLY the unmanaged resources, as it is expected that your managed `IDisposable`s will also do the same (and in fact you have no guarantee that they are still valid.  They may have finished being garbage collected already).  Calling `Dispose` will suppress the finalizer so that it doesn't waste time doing nothing and instead goes directly to collection.

== Layers of abstraction ==

Prior to this commit, the layers of abstraction between Core and .NET classes were as follows:

- P/Invoke bindings to the C API (In the static class NativeRaw, if the next layer is required, Native otherwise)
- (sometimes) static class Native which is responsible for UTF-8 string / byte array marshalling (The C API exposes this via the FLSlice family, so it needs some coaxing to turn into a C# `string` or `byte[]`)
- LiteCoreBridge utility class, which will check the out error of any API and throw an exception if it is non-zero
- C# class which interacts with LiteCore

This commit introduces a new layer exactly in the middle here (between Native and LiteCoreBridge) called NativeSafe.  This class will once again mirror the calls present in Native (or occassionally NativeRaw) except that in certain cases where thread safety is important, it will wrap the pointer objects that are returned in a wrapper class designed for that specific pointer (e.g. C4Database * -> C4DatabaseWrapper).  The reason it does this is so that the required locks for implementing thread safety can be stored right along with the pointers themselves.  Due to the unfortunate fact that .NET cannot make a generic type restricted to "blittable" (i.e. opaque types with only primitive members), there is a lot of repetition in the wrappers that would be reduced in the presence of a generic type NativeWrapper<T>.

== Process of Implementation ==

The process to get to this point was to go through the public LiteCore C headers one by one and look for functions that .NET uses directly.  If there are no functions present that .NET uses, or every one of them was marked thread safe for that file, then no action was taken.  Otherwise, a file called <header>_native_safe.cs (to parallel the existing <header>_native.cs autogenerated source) was created and EVERY function .NET uses was added to it, regardless of its thread safety status.  If the function is thread safe, then the NativeSafe class just takes the raw pointer out of the wrapper and uses it as-is for that one call.  If it needs any thread safety, the NativeSafe class will use varying implementations of `UseSafe` which will seize the appropriate locks, present the pointer to the caller in a callback, and then release all the locks in reverse order once the callback exits.  And so in this way, wrappers can be passed around and used with the NativeSafe API without any worrying about whether or not locks need to be seized by the user.

== Why do C# classes still have ThreadSafety though? ==

The ThreadSafety that is present in Database, Collection, and Scope (and maybe others) is there to eliminate races between various actions and closing a database.  The native wrappers do not account for this.  So in many cases, at least two locks will be seized in a given operation.  One to verify that a database is Open, and will stay that way for a while, and another to make sure that the thread safety of the native object is honored (even if it happens to be accessed by another C# class).

== Notes for each header ==

c4Base.h -> All thread-safe
c4BlobStore.h -> wrappers made for C4ReadStream and C4WriteStream, only instance locks
c4Certificate.h -> All thread-safe
c4Collection.h -> wrapper made for C4Collection (reuses the same lock from C4Database, has no need for instance locking)
c4Database.h -> wrapper made for C4Database, only instance lock
c4DocEnumerator.h -> all unused by .NET
c4Document.h -> Repurposed existing C4DocumentWrapper with instance lock and database lock to choose from (or both)
c4Document+Fleece.h -> Unusual FLEncoderWrapper made.  This class hides the native pointer, and holds a lock for its entire lifetime to implement the requirement that not only does the retrieval of the underlying FLEncoder need to be protected, but also its usage.  This type is only ever used inline, and should never be held across any function boundaries.  It is ONLY needed for the _shared_ database encoder.
c4Index.h -> wrappers made for C4Index and C4IndexUpdater (former only instance lock, latter also with Database lock)
c4Listener.h -> All thread-safe
c4Log.h -> All thread-safe (and not really relevant to state anyway)
c4Observer.h -> wrapper made for C4QueryObserver (reuses the same lock from C4Database, has no need for instance locking)
c4PredictiveQuery.h -> All thread-safe
c4Query.h -> wrappers made for C4Query and C4QueryEnumerator (former instance and Database lock, latter instance only)
c4Replicator.h -> wrapper made for C4Replicator, only instance lock (Most of these calls uses a state machine on the Core side which make them thread safe)
c4Socket.h -> wrapper made for C4Socket, only instance lock


bmeike (approved):
I'm really not equipped to read this whole thing.  It looks architecturally right.. though I have to say that using the scope of a variable to define the scope of a lock is slightly harder to read than a block.

bmeike:
... also, FWIW, the "all checks have failed" thing continues to be annoying.  If their failure is expected, perhaps best to disable them, so that significant failures are apparent?

borrrden:
In this case the failure is because of two PRs being dependent on each other.  I could remove all platforms except ones that reliably pass (i.e. Windows) but I have a feeling that I will simply never enable them again.

With regards to readability I much prefer the scoped version.  The block version comes with some restrictions that can get annoying regarding taking the address of variables that are used in both the inner and outer scope not being allowed, etc. 

Microsoft has somewhat authorized this pattern too with their .NET 9 inclusion of a similar looking lock https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock.enterscope?view=net-9.0

PR #1660, opened 2024-09-14
Fix up issues that probably came in during a rebase

borrrden:
There are a lot of things going on here but this gives me a green test run:

1. Incorrect logic for firing document changed listener
2. Revert to NQuery and XQuery that create the internal C4Query in a void method, rather than returning
3. Database test for query listener has redundant logic that seems to prove nothing
4. Migrate two tests that were testing two modes each into one test each with two variants to be more clear about which one failed
5. Fix a test issue comparing collections from different databases since that was changed to not be allowed
6. Change ports on TestBusyPort so that it doesn't fail when I happen to have Sync Gateway running (which uses the same port)
7. Fix awful error where nativewrapper will never free any native objects because it's ref count starts at 0 instead of 1
8. Retain C4Collection and C4Socket's that don't get created just for the platform side
9. Bump LiteCore to 3.3.0-45
10. Get rid of Jenkins dead weight that is not helping much.  This will give more green PR validations, but the post build tests will still be a pain to work with.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,47 +38,7 @@ pipeline {
                         }
                         stage(".NET 6 Windows") {
                             steps {
-                                catchError {
-                                    powershell 'jenkins\\run_net6_tests.ps1'
-                                }
-								
-                                echo currentBuild.result
-                            }
-                        }
-                    }
-                }
-	            stage("Mac Node") {
-		            agent { label 'dotnet-mobile-mac-mini'  }
-			        environment {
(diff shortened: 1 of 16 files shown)


bmeike (approved):
I'm not going to pretend to understand this.
I have two small comments:
- Is this just github formatting or is the indentation really wacko
- do you really need both the hashes and the version for LiteCore?

borrrden:
No I don't need both of the hashes, but I want them just in case I need to quickly check out the correct source

PR #1663, opened 2024-09-26
CBL-6253: Use lock when calling stream close methods

borrrden:


bmeike (approved):
yeah.... I cannot find any pattern in whether I have to lock or not, calling free/release

borrrden:
Very true...this came from a slack discussion and the headers have not been updated in core yet.

PR #1664, opened 2024-09-27
Cleanup defines in the project

borrrden:
Remove defunct WINDOWS_UWP and NETSTANDARD2_0
Bump to net8.0 console since net6.0 support is ending in Nov 2024


borrrden:
The tests didn't actually run here because the other repo needs to update the invocation of the tests to use net8.0 now.  But see the other PR for the run that happened.

bmeike (approved):

PR #1665, opened 2024-09-27
CBL-6236: Include the exe directory in the LiteCore search path

borrrden:
Self contained .NET applications will collapse into a single architecture, and will copy everything needed to the output directory automatically.  This means LiteCore.dll will be next to the executable and not in any subdirectory as it was in .net framework.


borrrden:
Test failure is Maui iOS (changes to only PR validate on Windows have not been backported to this branch yet) but Windows passed.

pasin (approved):

PR #1670, opened 2024-10-12
CBL-6236: Include the exe directory in the LiteCore search path (#1665)

borrrden:
Self contained .NET applications will collapse into a single architecture, and will copy everything needed to the output directory automatically.  This means LiteCore.dll will be next to the executable and not in any subdirectory as it was in .net framework.


borrrden:
Cherry picked from 1c873ee02d04c44c2975620120bcfd6f311b32f8

bmeike (approved):

PR #1671, opened 2024-10-21
CBL-6329: QueryBuilder queries should dispose child objects

borrrden:
This is slightly strange in that some of the child objects are actually the same object as the parent, so check for that as well.  Each chain in a querybuilder call chain basically creates a new object that wraps the old one and appends additional information to itself, but the appended information was not being properly disposed.

### DIFF
--- a/src/Couchbase.Lite.Shared/Query/XQuery.cs
+++ b/src/Couchbase.Lite.Shared/Query/XQuery.cs
@@ -70,18 +70,27 @@ namespace Couchbase.Lite.Internal.Query
 
         #region Protected Methods
 
-        protected void Copy(XQuery source)
+        protected void Move(XQuery source)
         {
+            // These things will be wrapped by this object now, 
+            // so the original object doesn't need to worry about
+            // the disposables anymore
             Database = source.Database;
             Collection = source.Collection;
             SelectImpl = source.SelectImpl;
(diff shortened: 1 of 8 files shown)


borrrden:
One test failure that is not related to this PR (has to do with TLS)

src/Couchbase.Lite.Shared/Query/XQuery.cs
```diff
@@ -70,18 +70,27 @@ internal class XQuery : QueryBase
 
         #region Protected Methods
 
-        protected void Copy(XQuery source)
+        protected void Move(XQuery source)
```
pasin, line 73: I don't think I have the full picture to understand what will happen. If I have something like this (ignore the exact type name and syntax) :

From from = select(...).from(DataSource.Collection(collection))

Where where 1 = from.where(....)
Where where 2 = from.where(....)

Will there where2 has a valid data source?

borrrden: Hmmm no, it won't, but this situation will never work here:

If I don't "move" then disposing where1 will invalidate where2.  If I do, then where2 cannot be created....

borrrden:
Incredible point made about how this affects users who save intermediate state.  Given that I don't think this can go in right now.

PR #1672, opened 2024-10-21
CBL-6348:  Implement RevisionIDs API (unsupported, for testing only)

borrrden:
This is just for parity so the testing server can build against both 3.2 and 4.0

### DIFF
--- a/src/LiteCore/src/LiteCore.Shared/Interop/C4Document.cs
+++ b/src/LiteCore/src/LiteCore.Shared/Interop/C4Document.cs
@@ -44,7 +44,7 @@ namespace LiteCore.Interop
 
     internal static unsafe partial class Native
     {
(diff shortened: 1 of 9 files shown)


src/LiteCore/src/LiteCore.Shared/Interop/C4Document.cs
```diff
@@ -41,6 +41,17 @@ internal unsafe partial struct C4Document
         public C4Revision selectedRev;
         public C4ExtraInfo extraInfo;
     }
+
+    internal static unsafe partial class Native
+    {
+        // This is only used for internal testing in version vector mode
```
borrrden, line 47: This comment is not correct for 3.2

pasin (approved):

pasin (approved):

PR #1678, opened 2024-11-13
Update the structs for LiteCore interop

borrrden:
Forgot to do this when upgrading LiteCore itself and it only manifest itself in some tests that are only run in release mode.

Nullability fix (Check() guarantees _c4Query is either null or an exception is thrown)

Fix a test for the new thread safety model (c4query_run is not blocked, just c4queryenum functionality on the query)

### DIFF
--- a/src/Couchbase.Lite.Shared/Query/XQuery.cs
+++ b/src/Couchbase.Lite.Shared/Query/XQuery.cs
@@ -120,7 +120,7 @@ namespace Couchbase.Lite.Internal.Query
                 }
 
                 Check();
-                return NativeSafe.c4query_run(_c4Query, (FLSlice)paramJson, err);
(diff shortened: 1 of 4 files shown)


src/Couchbase.Lite.Shared/Query/XQuery.cs
```diff
@@ -120,7 +120,7 @@ public override unsafe IResultSet Execute()
                 }
 
                 Check();
```
bmeike, line 122: Is this because Check() guarantees not null?  If so, could Check() return the non-null query?

bmeike (approved):

src/Couchbase.Lite.Shared/Query/XQuery.cs
borrrden: I had tried to do that previously but it ended up being too awkward for reasons I can't remember.  I should try again at some point.

PR #1681, opened 2024-11-20
CBL-6325: Implement MmapEnabled property on DatabaseConfiguration

borrrden:
Also

- Support for tool based test spec verification (ImplementsTestSpecAttribute)

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Database/Database.cs
+++ b/src/Couchbase.Lite.Shared/API/Database/Database.cs
@@ -1247,6 +1247,10 @@ namespace Couchbase.Lite
             //    config.flags |= C4DatabaseFlags.DiskSyncFull;
             //}
 
+            if(!Config.MmapEnabled) {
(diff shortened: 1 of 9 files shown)


borrrden:
This cherry pick applied cleanly, it should look virtually identical to the original 3.2 PR

borrrden:
There were a few cleanups needed from the cherry pick:

- Consolidate the definition of FLSliceResult in Fleece.cs (it had two different partial defs, so I made it into just one)
- Fix MmapTest to account for the new thread safety API (need to use RawDatabase)

borrrden:
The same unrelated test is failing.  Something is wrong with the machine I think, making it so slow as to time out on that test.  But I don't think I have an alternative machine to run on.

pasin (approved):
Noticed that `config.flags |= C4DatabaseFlags.DiskSyncFull;` is commented out.

borrrden:
Good catch.  I have a compiler warning in place to remind myself to fix that later, when I go through the compiler warnings.

PR #1685, opened 2024-12-07
CBL-6190 CBL-6191: API for array indexing

borrrden:
Cherry picked from b9cc9e363890ca0a2a4c528acde459c901cc3110

> Conflicts:
	Jenkinsfile
	core_version.ini
        src/Couchbase.Lite.Shared/API/Database/Collection.cs

Jenkinsfile conflicts were from changes to the git cloning process (using pre-4.0 branch) and the name of the powershell script, fixed easily resulting in only a change to the name of the step core_version conflicts as usual because it is also changed on the other branch (it remains unchanged here) Collection.cs conflicts due to the new locking strategy, again the result is no change on this branch.

There was also an issue with CreateIndex not getting the new "N1QL only" variant (removed IndexConfiguration.QueryLanguage since it was always N1QL anyway).
And an issue in ArrayIndexTest where the raw pointer needs to be used instead of the new threadsafe wrapper (TestNativeRaw.c4coll_getIndexesInfo(collection.c4coll -> TestNativeRaw.c4coll_getIndexesInfo(collection.c4coll.RawCollection))


src/Couchbase.Lite.Shared/Query/IndexConfiguration.cs
```diff
@@ -28,53 +31,52 @@ namespace Couchbase.Lite.Internal.Query
...
-            : this(indexType, C4QueryLanguage.N1QLQuery)
-        {
-            Expressions = items;
-        }
-
-        internal IndexConfiguration(C4IndexType indexType, C4QueryLanguage queryLanguage)
+        internal IndexConfiguration(C4IndexType indexType, string[]? expressions)
         {
+            if(indexType == C4IndexType.ArrayIndex && expressions != null) {
+                // Quick sanity check.  If I allow this to proceed, no error will happen but
```
bmeike, line 48: There was an extended discussion with Jens, about what, exactly, we should check, here.  His position was that we should leave as much as possible to LiteCore; that the platforms should do as little checking as possible.
The conclusion was that the platform APIs should follow these rules:
- If the client code passes `null` as the expression list, we should pass an empty string to LiteCore as the expression spec
- It is an error for the client code to pass a singleton list containing null or the empty string
- All other espression specs should be passed, comma separated, to LiteCore, for checking.  In particular, an empty list should be legal (it is a synonym for passing null).

borrrden: Okey dokey guess I'll be silently changing the behavior for 4.0 then.

bmeike (approved):
Thanks for the complete cherry-pick notes.  Much appreciated.

PR #1686, opened 2024-12-13
CBL-6540: AllowReplicatingInBackground behavior is opposite of what is documented

borrrden:
The replicator automatically stop when going in the background and the value of the property is false (the default).  Currently, it is reversed and only automatically stops when the property is true.


bmeike (commented):
This is a breaking change.  Should we really be doing it in a point release?

It will silently break customer code.

borrrden:
Indeed I pointed this out before and then a ticket got filed anyway so I did it.

bmeike (approved):

pasin (approved):
PM approved the breaking change and will work closely with the customer regarding the change.

PR #1688, opened 2024-12-18
CBL-6482: Create two variants (packaged / unpackaged) of defaultdirectoryresolver

borrrden:
Also evaluate it lazily to save some overhead

Cherry picked cleanly from 97bdb27112a0727248ac0f3d12cdfeb3876eecbb


borrrden:
Test failures are addressed in another PR and are not related to this.

PR #1690, opened 2024-12-20
Fix compilation error for .NET iOS

borrrden:
Add more steps to PR validation to catch this issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,6 +36,16 @@ pipeline {
                                 '''
                             }
                         }
+                        stage("Compile CE") {
+                            steps {
+                                powershell 'jenkins\\compile_ce.ps1'
+                            }
+                        }
+                        stage("Compile EE") {
+                            steps {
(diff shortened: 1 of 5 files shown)


borrrden:
Note: depends on https://github.com/couchbaselabs/couchbase-lite-net-ee/pull/194

bmeike (approved):
Don't quite understand this change.  Can't review it carefully.

borrrden:
My bad I forgot to include an explanation.  Essentially every time I make a cherry pick from 3.x these days I have to refactor calls to LiteCore to use the new thread safety layer (NativeSafe).  I forgot to do that in a previous commit and it wasn't caught because the only configuration that gets built is Windows and this was in iOS only code.  Now the PR validation will compile all platforms for both CE and EE.

borrrden:
The "JUST_CSHARP" stuff is so that I can compile all of the C# without needlessly downloading LiteCore.

PR #1693, opened 2025-01-16
Update LiteCore to 4.0.0-37

borrrden:


borrrden:
This LiteCore version is rejected

PR #1696, opened 2025-01-31
Fix a bizarre native AOT problem with iOS

borrrden:
Without doing this, accessing Database.Log throws a null reference exception (or some other nonsense).

Fix some incorrect comments as well.  Note that trying to do the log initialization using atomic swaps causes a hang, and I don't think it's enough of an issue to figure out why so just do a noobie swap.


pasin (approved):
It works before and then stops working?

borrrden:
I tried using a lock but that caused other issues. This started happening after the new logging stuff went in so the answer to why is not straightforward. I’m hoping that once the old api is removed I can put it back to the way it was.

PR #1700, opened 2025-03-04
CBL-6798: Don't allocate memory for c4log_getDomain

borrrden:
Now that the C API copies the domain name on input, it is no longer important for the C# side string to live pinned past the method invocation.


bmeike (approved):
Yeah... This vanished into git history, somewhere, for Java.  None of the branch heads still have the code that did it.  I *know* it was there at some point.  Glad it is gone.

pasin (approved):

PR #1702, opened 2025-04-15
Update LiteCore to 3.2.3-8

borrrden:


pasin (approved):

bmeike (approved):
The C4KeyPair stuff is because LiteCore changed the struct?

borrrden:
> The C4KeyPair stuff is because LiteCore changed the struct?

Yes. C# must mirror the struct because that’s how it marshals memory between managed and unmanaged. If it doesn’t know about the extra pointer at the end it will just be garbage instead of null.

PR #1713, opened 2025-06-25
CBL-7052: Completely remove FluentAssertions from testing

borrrden:
On the way I was able to fix a few things:

- Get rid of all WaitHandle.WaitAll calls so that the .NET 4.6.2 tests stop failing because xUnit runs on STAThread threads.
- Turn off version vectors since this is 3.3.0 now


bmeike (approved):
I don't understand the check failures.... but this looks good to me.

PR #1715, opened 2025-06-26
CBL-7111: Correct bug where partial indexes don't function in CE

borrrden:
The native struct layout differs depending on EE vs CE and that needs to be reflected.  The actual change here is in C4IndexTypes_defs.cs, and the rest is stuff that the bindings script picked up on its run.  I also took this chance to output some more info in the generate bindings script to warn me about which structs I should check when updating the bindings.


bmeike (approved):
Is there any way to warn if those things don't agree??

borrrden:
I would love it if there were but C# is set up to be willfully ignorant of the underlying native implementation.  I suppose it is akin to the relationship between `native` Java methods and their JNI implementations.

PR #1723, opened 2025-08-28
CBL-7380: Remove MmapEnabled

borrrden:
It is no longer applicable since it is always off

Also remove some more defunct default values


borrrden:
Test failures here unrelated to the PR

PR #1728, opened 2025-09-04
Overhaul for platform identification

borrrden:


pasin (approved):
I saw some are switched from the original __IOS__ to APPLE. I can't tell of all are correct or not.

PR #1729, opened 2025-09-05
Migrate off some third party dependencies

borrrden:
SimpleInjector -> Microsoft.Extensions.DependencyInjection Newtonsoft.Json -> System.Text.Json

Also remove DerConverter because it is not used for anything anymore.

This requires rewriting a bunch of code to use the new dependencies but it's largely similar, just with different naming


borrrden:
Also, removed all the code that is never compiled and hidden behind CBL_LINQ.  It's unlikely that the feature will see the light of day any time soon and by then it will be uncompilable

pasin (approved):

PR #1733, opened 2025-09-19
Catch task canceled exception when waiting for conflicts

borrrden:
Otherwise it could bring down the process when the replicator stops


borrrden:
The failing test is not related to this but I'm pretty sure I figured out why it fails and can fix it in the next PR.

PR #1747, opened 2026-02-17
CBL-7873: iOS / Mac Catalyst PAC file handling

borrrden:
Ported from lite-ios and I incidentally discovered that with MAUI I don't need to do all this P/Invoke nonsense, the bindings are essentially right there for me (though they are strangely in the CoreFoundation.CFNetwork namespace rather than just CFNetwork where I'd expect them).  I did not port this to our unsupported .NET macOS desktop platform, only .NET MAUI.  PAC files will continue to cause problems in macOS desktop.


pasin (approved):
Sorry, I have just seen this PR.

PR #1754, opened 2026-05-01
Bump LiteCore to 3.2.6-1

borrrden:


borrrden:
This is a fix that was meant for 3.2.5 but didn't make it in time, and so was reverted until we could work on it some more.  A backport from a reported issue with 3.3 on other platforms.

pasin (approved):

PR #1757, opened 2026-05-12
CBL-8114: Have replicator filter error conditions return false

borrrden:
This prevents unintended document leakage from filter errors or product bugs

### DIFF
--- a/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
+++ b/src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
@@ -510,10 +510,11 @@ public sealed unsafe class Replicator
 #endif
     private static bool PullValidateCallback(C4CollectionSpec collectionSpec, FLSlice docID, FLSlice revID, C4RevisionFlags revisionFlags, FLDict* dict, void* context)
     {
-        var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
-        if (replicator == null) {
-            WriteLog.To.Database.E(Tag, "Pull filter context pointing to invalid null replicator, aborting and returning true...");
-            return true;
+        var gcHandle = GCHandle.FromIntPtr((IntPtr)context);
+        if (gcHandle.Target is not Replicator replicator) {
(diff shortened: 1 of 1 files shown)


pasin (approved):

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
```diff
@@ -510,10 +510,11 @@ internal void CheckForCookiesToSet(WebSocketWrapper wrapper)
...
     private static bool PullValidateCallback(C4CollectionSpec collectionSpec, FLSlice docID, FLSlice revID, C4RevisionFlags revisionFlags, FLDict* dict, void* context)
     {
-        var replicator = GCHandle.FromIntPtr((IntPtr)context).Target as Replicator;
-        if (replicator == null) {
-            WriteLog.To.Database.E(Tag, "Pull filter context pointing to invalid null replicator, aborting and returning true...");
-            return true;
+        var gcHandle = GCHandle.FromIntPtr((IntPtr)context);
+        if (gcHandle.Target is not Replicator replicator) {
+            WriteLog.To.Database.E(Tag, "Pull filter context pointing to invalid null replicator, aborting and returning false...");
+            WriteLog.To.Database.E(Tag, $"Allocated: {gcHandle.IsAllocated}, Target Type: {gcHandle.Target.GetType().Name}");
```
pasin, line 516: Could `gcHandle.Target.GetType().Name` cause null pointer exception?

pasin (commented):
Just notice that after approving the PR.

pasin:
I couldn't think how all those cases could happen, but yes, for security reason, false should be returned.

src/Couchbase.Lite.Shared/API/Sync/Replicator.cs
borrrden: Apparently not.  It is not marked as nullable.

borrrden:
They are purely based on attributes so that the compiler is satisfied in some cases.  I don't think they will realistically happen either.

PR #1766, opened 2026-06-04
Add back 3.x API

borrrden:
The least messy way to handle this situation, in which we want a 4.1 feature in 3.4, is to take the 4.1 build and re-add all the removed API from it (as well as attempt to take out any irrelevant 4.x API).  The LiteCore build is identical except that it has restored rev-tree mode.  Also many tests had to be found and restored.


borrrden:
As noted, this is the strategy chosen for rolling out the 3.4 release and was agreed upon in a sync up meeting.

pasin (approved):
LGTM

PR #1767, opened 2026-06-05
CBL-8393: Updated conflict resolution logic

borrrden:
1. The local revid can be used as the winner now in local-wins cases
2. A merge body is only created when there is an actual merge


borrrden:
This is a fix that was accidentally dropped from 4.0.x, hence its inclusion

pasin (approved):

PR #1769, opened 2026-06-08
Use revtrees as is required for 3.x

borrrden:


borrrden:
Need to revert back to 3.x behavior, this is part of that

pasin (approved):

PR #1771, opened 2026-06-10
Ported crash and reachability fix from 4.1

borrrden:
Clean cherry pick from bf51bafaed49ba48057754eb43db50762d3847e8


borrrden:
Port fix from 4.1 that causes iOS replication to be unusable

pasin (approved):

PR #1779, opened 2026-07-06
Support Native AOT and trimming: remove reflection from JSON serialization and DI

jormenjanssen:
## Problem

Applications that publish with .NET Native AOT or trimming (`PublishAot` / `PublishTrimmed`) cannot currently use Couchbase Lite .NET. The library depends on runtime reflection in several places, which the trimmer strips out and Native AOT cannot compile:

- **The query path builds its JSON with the reflection-based `System.Text.Json` serializer.** `XQuery.EncodeAsJSON()` (the JSON handed to `c4query_new2`), `Collection.CreateIndex`, `Result.ToJSON()`, and `Parameters` (whose JSON feeds `c4query_setParameters` for live queries) all call `JsonSerializer.Serialize` with the default reflection resolver. The document `SetJSON` / `ToJSON` paths do the same for deserialization.
- **Dependency injection scans the assembly.** `Service` used `Assembly.GetTypes()`, attribute reflection, and `Activator.CreateInstance` to find `[CouchbaseDependency]` implementations.
- **`FLValueConverter` creates containers dynamically** via `Activator.CreateInstance(typeof(List<>).MakeGenericType(...))`.
- **`LoadLiteCore` locates the native library through `Assembly.Location`**, which returns an empty string in single-file and Native AOT apps, so LiteCore can never be found in those deployment models.

With the trim/single-file/AOT analyzers enabled, the library produces **102 warnings** (IL2026, IL2070, IL2072, IL2075, IL3000, IL3050). After this PR it produces **zero**, and a Native AOT console app can use the database, documents, blobs, indexes, and both QueryBuilder and SQL++ queries.

## What changed

**JSON is now written by hand and read through source generation.** Couchbase Lite only ever serializes a closed set of value types (null, booleans, numbers, strings, dates, binary data, dictionaries, lists, `Blob`, `DictionaryObject`/`ArrayObject`), so a new internal `CouchbaseJson` class writes those directly with `Utf8JsonWriter` — no reflection, byte-identical output for the types LiteCore parses. For reading, a source-generated `JsonSerializerContext` (`CouchbaseJsonContext`) is installed as the `TypeInfoResolver` on the shared serializer options, and `DataOps.ParseTo<T>` uses the AOT-safe `JsonTypeInfo` overload. The three custom converters (`CouchbaseConverter`, `IArrayConverter`, `IDictionaryObjectConverter`) are now fully manual in both directions. `JsonElement` values are converted by walking the element tree instead of re-entering the serializer.

**Dependency injection registers explicitly.** Which implementations exist per platform is already decided at compile time by the `CBL_PLATFORM_*` defines, so each platform block in `Service` now registers its implementations directly, with the same lifetimes the `[CouchbaseDependency]` attributes declared. The attributes stay in place as public API and documentation.

**`FLValueConverter` materializes `List<object>` / `Dictionary<string, object>` directly.** The removed `hintType` parameter only shaped the generic type of the top-level container; no library code passed one, and the single test that did now casts per element.

**`LoadLiteCore` falls back to `AppContext.BaseDirectory`** when `Assembly.Location` is empty, which is where the native libraries live in single-file and AOT published apps.

**Analyzers stay on.** `IsAotCompatible=true` is set for the `net8.0` target of `Couchbase.Lite` and `Couchbase.Lite.Support.NetDesktop`, so regressions surface as build warnings and the published packages carry the `IsTrimmable` metadata.

### Bugs found along the way

- The Fleece encoder's `JsonElement` array/object branches called `FLEncode()` without passing the target encoder, discarding the encoded result instead of writing it into the document. Fixed by the tree-walking conversion.
- `IArrayConverter` deserialized `Blob` through its public properties, which cannot work (`Blob` has multiple public constructors and no parameterless one, so `JsonSerializer` throws). It now uses the validating `Blob(IDictionary)` constructor.
- `Parameters.ToString()` serialized `DictionaryObject`/`ArrayObject` values through their public surface, producing malformed parameter JSON for live queries. They are now written as proper JSON objects/arrays, matching what the Fleece encoding of the same parameters produces at `Execute()` time.

## Verification

- `Couchbase.Lite` (net8.0 and net462) builds with **0 warnings** with the trim, single-file, and AOT analyzers enabled (down from 102).
- Verified with a local console harness referencing the local projects (not included in this PR), running with `JsonSerializerIsReflectionEnabledByDefault=false` so that any missed reflection-based JSON call site throws immediately. All steps pass on JIT and as a fully trimmed self-contained publish: open database, CRUD with nested values and a blob, `ToJSON`/`SetJSON` round trip, `Blob.ToJSON`, value index creation, QueryBuilder query with parameters, SQL++ query, `Result.ToJSON`/`ToDictionary`.
- Reverse P/Invoke was audited: all native callbacks are concrete delegate types annotated with `[UnmanagedFunctionPointer(CallingConvention.Cdecl)]` and passed statically in the binding signatures, which Native AOT supports with compile-time thunks.

## Notes for maintainers

- The Enterprise Edition compiles these same shared sources; EE-only (`#if COUCHBASE_ENTERPRISE`) code that uses reflection-based JSON will want the same treatment. `CouchbaseJsonContext` registrations can be extended if EE deserializes additional root types.
- `FLValueConverter.ToCouchbaseObject` lost its internal `hintType` parameter; if EE passes one, the fix is mechanical and surfaces at compile time.
- The MAUI targets were deliberately left out of `IsAotCompatible` for now — the mobile AOT pipelines have different constraints and seem better handled as a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Couchbase.Lite.Shared/Document/DataOps.cs
+++ b/src/Couchbase.Lite.Shared/Document/DataOps.cs
@@ -56,9 +56,11 @@ internal static class DataOps
         T? retVal;
         try {
             // The JsonTypeInfo overload is AOT safe; T must be registered on
-            // CouchbaseJsonContext (or handled by a registered converter)
+            // CouchbaseJsonContext (or handled by a registered converter).
+            // Only JsonException maps to InvalidJSON so that registration
+            // failures (NotSupportedException etc.) surface immediately.
             retVal = (T?)JsonSerializer.Deserialize(json, SerializerOptions.GetTypeInfo(typeof(T)));
-        } catch {
+        } catch (JsonException) {
             throw new CouchbaseLiteException(C4ErrorCode.InvalidParameter, CouchbaseLiteErrorMessage.InvalidJSON);
         }
 
--- a/src/Couchbase.Lite.Shared/Document/DataOps.cs
+++ b/src/Couchbase.Lite.Shared/Document/DataOps.cs
@@ -30,62 +30,12 @@ internal sealed class CouchbaseConverter : JsonConverter<object>
 {
     public override object? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        return reader.TokenType switch
-        {
-            JsonTokenType.True => true,
-            JsonTokenType.False => false,
-            JsonTokenType.Number => ReadNumber(ref reader),
-            JsonTokenType.String => reader.GetString(),
-            JsonTokenType.StartObject => ReadObject(ref reader),
-            JsonTokenType.StartArray => ReadArray(ref reader),
-            JsonTokenType.Null => null,
-            _ => throw new JsonException($"Unexpected token parsing JSON. Token: {reader.TokenType}")
-        };
+        return CouchbaseJson.ReadValue(ref reader);
     }
 
     public override void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options)
     {
-        JsonSerializer.Serialize(writer, value, options);
-    }
-    
-    private object ReadNumber(ref Utf8JsonReader reader)
-    {
-        if (reader.TryGetInt64(out var l))
-            return l;
-
-        return reader.GetDouble();
-    }
-
-    private Dictionary<string, object> ReadObject(ref Utf8JsonReader reader)
-    {
-        var dictionary = new Dictionary<string, object>();
-        
-        while (reader.Read()) {
-            if (reader.TokenType == JsonTokenType.EndObject)
-                break;
-
-            if (reader.TokenType != JsonTokenType.PropertyName)
-                throw new JsonException("Expected property name");
-
-            var propertyName = reader.GetString()!;
-            reader.Read();
-            dictionary[propertyName] = Read(ref reader, typeof(object), null!)!;
-        }
-
-        return dictionary;
-    }
-    
-    private List<object> ReadArray(ref Utf8JsonReader reader)
-    {
-        var list = new List<object>();
-        while (reader.Read()) {
-            if (reader.TokenType == JsonTokenType.EndArray)
-                break;
-
-            list.Add(Read(ref reader, typeof(object), null!)!);
-        }
-
-        return list;
+        CouchbaseJson.WriteValue(writer, value);
     }
 }
 
@@ -93,15 +43,21 @@ internal static class DataOps
 {
     internal static readonly JsonSerializerOptions SerializerOptions = new()
     {
+        // The source generated resolver keeps System.Text.Json off of its
+        // reflection based code paths so that deserialization works under
+        // Native AOT / trimming.  The converters do all of the actual work.
+        TypeInfoResolver = CouchbaseJsonContext.Default,
         Converters = { new CouchbaseConverter(), new IArrayConverter(), new IDictionaryObjectConverter() },
         PreferredObjectCreationHandling = JsonObjectCreationHandling.Replace
     };
-    
+
     internal static T? ParseTo<T>(string json)
     {
         T? retVal;
         try {
-            retVal = JsonSerializer.Deserialize<T>(json, SerializerOptions);
+            // The JsonTypeInfo overload is AOT safe; T must be registered on
+            // CouchbaseJsonContext (or handled by a registered converter)
+            retVal = (T?)JsonSerializer.Deserialize(json, SerializerOptions.GetTypeInfo(typeof(T)));
         } catch {
             throw new CouchbaseLiteException(C4ErrorCode.InvalidParameter, CouchbaseLiteErrorMessage.InvalidJSON);
         }
@@ -186,23 +142,7 @@ internal static class DataOps
             case ArrayObject roarr and not MutableArrayObject:
                 return roarr.ToMutable();
             case JsonElement jobj:
-                switch (jobj.ValueKind) {
-                    case JsonValueKind.Array:
-                        return ConvertList(jobj.Deserialize<IList>(SerializerOptions)!);
-                    case JsonValueKind.Object:
-                        return ConvertDictionary(jobj.Deserialize<IDictionary<string, object>>(SerializerOptions)!);
-                    case JsonValueKind.Number:
-                        return jobj.TryGetInt64(out var l) ? l : jobj.GetDouble();
-                    case JsonValueKind.String:
-                        return jobj.GetString();
-                    case JsonValueKind.True:
-                    case JsonValueKind.False:
-                        return jobj.GetBoolean();
-                    case JsonValueKind.Null:
-                        return null;
-                    default:
-                        throw new ArgumentException("Invalid JsonElement type: " + jobj.ValueKind);
-                }
+                return ToCouchbaseObject(CouchbaseJson.ToNetObject(jobj));
             case IDictionary<string, object?> dict:
                 return ConvertDictionary(dict);
             case IList list:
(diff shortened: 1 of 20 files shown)


jormenjanssen:
I have signed the Couchbase New Contributor Agreement (as j.janssen@riwo.eu, matching the commit author email on this PR).

borrrden:
Oh this is juicy!  This is an unrealized goal that I've had for a while so I'm happy someone has taken an attempt at it.  It's not even terribly big which means it should be reviewable in a fair amount of time.  Thanks!

src/Couchbase.Lite.Shared/Serialization/CouchbaseJson.cs
```diff
@@ -0,0 +1,265 @@
...
+// DataOps.SerializerOptions, which are entirely hand written.
+[JsonSourceGenerationOptions(GenerationMode = JsonSourceGenerationMode.Metadata)]
+[JsonSerializable(typeof(object))]
+[JsonSerializable(typeof(Dictionary<string, object>))]
+[JsonSerializable(typeof(IDictionary<string, object>))]
+[JsonSerializable(typeof(List<object>))]
+[JsonSerializable(typeof(IList<object>))]
+[JsonSerializable(typeof(IList))]
+[JsonSerializable(typeof(IList<IList<object>>))]
+internal sealed partial class CouchbaseJsonContext : JsonSerializerContext
```
borrrden, line 46: Need one more (from EE) entry here 

```suggestion
[JsonSerializable(typeof(Dictionary<string,string[]>))]
internal sealed partial class CouchbaseJsonContext : JsonSerializerContext
```

src/Couchbase.Lite.Shared/Document/DataOps.cs
```diff
@@ -30,78 +30,34 @@ internal sealed class CouchbaseConverter : JsonConverter<object>
...
+
     internal static T? ParseTo<T>(string json)
     {
         T? retVal;
         try {
-            retVal = JsonSerializer.Deserialize<T>(json, SerializerOptions);
+            // The JsonTypeInfo overload is AOT safe; T must be registered on
+            // CouchbaseJsonContext (or handled by a registered converter)
+            retVal = (T?)JsonSerializer.Deserialize(json, SerializerOptions.GetTypeInfo(typeof(T)));
         } catch {
```
borrrden: This is going to be too broad now.  Let's catch just `JsonException` so that actual errors in JSON will return InvalidJson, and the fallback "not supported" and other errors from registration failures will get realized sooner.

src/Couchbase.Lite.Shared/Serialization/CouchbaseJson.cs
```diff
@@ -0,0 +1,265 @@
...
+// Reflection free JSON reading and writing for the closed set of value types that
+// Couchbase Lite supports (see DataOps.ToCouchbaseObject).  These methods replace
+// the reflection based JsonSerializer code paths, which are unavailable under
+// Native AOT and produce trim warnings (IL2026 / IL3050).
+internal static class CouchbaseJson
+{
+    // Serializes a tree of Couchbase supported values to a JSON string.
+    internal static string Serialize(object? value)
+    {
+        using var stream = new MemoryStream();
```
borrrden, line 68: This is allocation heavy in a path that is likely to be hit a lot.  It would be better to make use of `ArrayBufferWriter<byte>` here (bonus is that no more `ToArray()` call will be needed).

borrrden (changes requested):
Just a few things I'd like to see changed

src/Couchbase.Lite.Shared/Serialization/CouchbaseJson.cs
jormenjanssen: Added in 3d803875, exactly as suggested.

src/Couchbase.Lite.Shared/Document/DataOps.cs
jormenjanssen: Done in 6dc7a3e1 - only `JsonException` maps to InvalidJSON now, so `NotSupportedException` from an unregistered root type surfaces immediately instead of being disguised as a JSON error.

src/Couchbase.Lite.Shared/Serialization/CouchbaseJson.cs
jormenjanssen: Done in 404ad7a4 - `ArrayBufferWriter<byte>` + `Encoding.UTF8.GetString(WrittenSpan)` on .NET 6+. The net462 build keeps the `MemoryStream` path since `ArrayBufferWriter` is not available on .NET Framework.

Measured with BenchmarkDotNet (`MemoryDiagnoser`, net10.0/win-x64), both variants using the identical `WriteValue` walker so only the buffer strategy differs:

| Method | Mean | Allocated | Alloc Ratio |
|---|---:|---:|---:|
| QueryTree_MemoryStream | 518.2 ns | 1.76 KB | 1.00 |
| QueryTree_ArrayBufferWriter | 477.2 ns | 1.25 KB | 0.71 |
| LargeDoc_MemoryStream | 16,686.9 ns | 50.78 KB (Gen1 promotions) | 28.89 |
| LargeDoc_ArrayBufferWriter | 16,111.2 ns | 32.31 KB (no Gen1) | 18.38 |

~29% fewer bytes allocated for a typical query tree and ~36% for a larger document, and the Gen1 promotions from the `ToArray()` copy are gone. The benchmarks plus a set of serialization unit tests (reflection-parity, deep-nesting round trips) live on a testing branch on my fork to keep this PR lean: https://github.com/jormenjanssen/couchbase-lite-net/tree/feature/native-aot-json-sourcegen-tests

src/Couchbase.Lite.Support.NetDesktop/Couchbase.Lite.Support.NetDesktop.csproj
```diff
@@ -15,6 +15,10 @@
 	<SingleProject>true</SingleProject>
     <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
   </PropertyGroup>
+  <!-- Native AOT / trimming support.  Only applicable to .NET 7+. -->
+  <PropertyGroup Condition=" '$(TargetFramework)' == '$(ConsoleDotNetVersion)' ">
```
borrrden: I'm not sure what your intent here is.  There are several other monikers here that are also .NET 7+ (currently net8.0, and net10.0 for maui targets).  Do they all need to be considered here?

borrrden: (commented on the wrong one, this one only has desktop, but the OTHER one like this in the diff has more)

jormenjanssen: Good point - the intent was to enable the trim/AOT analyzers and ship the `IsTrimmable`/AOT metadata, and the net8.0-only condition was needlessly narrow. Replaced it with the documented pattern for libraries that multi-target frameworks without AOT analysis support: `$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))` - see the [Target framework requirements](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/#target-framework-requirements) section of the Native AOT docs. All net8.0+ TFMs (including the MAUI and WinUI targets) now carry `IsAotCompatible`; only net462 is excluded, where the property does not exist.

Verified locally with the MAUI workloads installed: all six TFMs build with **zero** trim/AOT analyzer warnings. The only warnings on the mobile/WinUI TFMs are pre-existing CA1416 platform-compatibility ones, unrelated to this change. Commit: 85f229bc

borrrden (approved):

PR #1791, opened 2026-09-11
Migrate to xUnit v3

borrrden:
v2 is now discontinued and also on the EE side move to a different runner harness since the process now is too flaky.

This PR also bumps .NET Framework 4.6.2 -> 4.7.2 for building since xUnit v3 requires that anyway, and it doesn't affect what we are supporting (Windows 11 ships with 4.8 or 4.8.1).  

There is no functional change to the product here, only testing related changes.


pasin (approved):
As PR validation passes, I assume that it's all correct.

PR #1794, opened 2026-09-16
CBL-8931: Lock access on c4db_release

borrrden:
Just in case something else is already using the native handle when this gets encountered.  Now it will wait for that to finish before freeing it.


borrrden:
Backport of CBL-8912

pasin (approved):

PR #1799, opened 2026-09-17
CBL-8283: Catch ObjectDisposedException in PerformRead 

borrrden:
It has the same reasoning as the other two that are already caught.  Shutting down the connection disposes the underlying streams.


borrrden:
Backport of CBL-8282

pasin (approved):

PR #1801, opened 2026-09-18
CBL-7874: iOS / Mac Catalyst PAC file handling

borrrden:
Turns out there a a bunch of runtime provided methods to do the stuff I was doing, so make use of them and add in PAC processing for both URL and script based PAC.


borrrden:
Backport of CBL-7873

